PR #1, opened 2022-09-29
tweak: remove cid column

Waves966:
## Description

It doesn't fix an issue but removed an unused column from the database.

## Checklist

I didn't load the server setter shit because im lazy.

- [x ] I have personally loaded this code into an updated qbcore project and checked all of its functionality.
- [ x] My code fits the style guidelines.
- [ x] My PR fits the contribution guidelines.

### DIFF
--- a/qbcore.sql
+++ b/qbcore.sql
@@ -1,7 +1,6 @@
 CREATE TABLE IF NOT EXISTS `players` (
   `id` int(11) NOT NULL AUTO_INCREMENT,
   `citizenid` varchar(50) NOT NULL,
-  `cid` int(11) DEFAULT NULL,
   `license` varchar(255) NOT NULL,
   `name` varchar(255) NOT NULL,
   `money` text NOT NULL,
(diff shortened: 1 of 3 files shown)


JericoFX:
cid is use for the multicharacter, cid is actually the amount of characters you have, works like the position of the caracter 1,2,3...
https://github.com/QBCore-Remastered/qb-multicharacter/blob/main/server/main.lua#L103
https://github.com/QBCore-Remastered/qb-multicharacter/blob/main/html/index.html#L61

Waves966:
Please view my qb-multicharacter PR which removes the new for the CID
column in multicharacter as well.

JericoFX:
im sorry now i understand what you want to achive, you move the cid inside the character

Waves966:
> 

Well not exactly. the cid is **already** in the charinfo is has since qbcore used steam

Manason:
> > 
> 
> Well not exactly. the cid is **already** in the charinfo is has since qbcore used steam

You are saying the player.cid field isn't actually being used? I would think we should prefer descriptive columns over JSON columns to have a stronger data model.

Waves966:
The whole player data needs to be reworked but that's not gonna happen so
here we are

PR #7, opened 2022-10-03
Basic Readibility for functions.lua

HebiKotei:
🎨 : ## Description

This pull request reworks some nested requirements that impede the readability of the ```functions.lua``` file. Functionality should remain exactly the same as the basic logic has not been changed. Nesting has been improved by using more guard clauses and removing else statements.

This is just a proof of concept for the moment, am unable to personally test due to my VM currently not being set up.

## Checklist

- [ ] I have personally loaded this code into an updated qbcore project and checked all of its functionality.

### DIFF
--- a/server/functions.lua
+++ b/server/functions.lua
@@ -117,55 +117,53 @@ end
 
 -- Will set the provided player id / source into the provided bucket id
 function QBCore.Functions.SetPlayerBucket(source --[[ int ]], bucket --[[ int ]])
-    if source and bucket then
-        local plicense = QBCore.Functions.GetIdentifier(source, 'license')
-        SetPlayerRoutingBucket(source, bucket)
-        QBCore.Player_Buckets[plicense] = {id = source, bucket = bucket}
-        return true
-    else
-        return false
-    end
+    if not (source or bucket) then return false end
+    
+    local plicense = QBCore.Functions.GetIdentifier(source, 'license')
+    SetPlayerRoutingBucket(source, bucket)
+    QBCore.Player_Buckets[plicense] = {id = source, bucket = bucket}
+    return true
 end
 
 -- Will set any entity into the provided bucket, for example peds / vehicles / props / etc.
 function QBCore.Functions.SetEntityBucket(entity --[[ int ]], bucket --[[ int ]])
-    if entity and bucket then
-        SetEntityRoutingBucket(entity, bucket)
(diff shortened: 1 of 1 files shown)


Daze13 (approved):
I have tested this pr on my local and it works as intended

server/functions.lua
```diff
@@ -315,18 +313,19 @@ function QBCore.Functions.AddPermission(source, permission)
...
                 ExecuteCommand(('remove_principal player.%s qbcore.%s'):format(source, v))
                 QBCore.Commands.Refresh(source)
             end
         end
+        return
+    end
+    
+    if IsPlayerAceAllowed(source, permission) then
+        ExecuteCommand(('remove_principal player.%s qbcore.%s'):format(source, permission))
+        QBCore.Commands.Refresh(source)
```
Manason, line 328: I think this function is less readable now.

server/functions.lua
```diff
@@ -335,7 +334,9 @@ end
 function QBCore.Functions.HasPermission(source, permission)
     if type(permission) == "string" then
         if IsPlayerAceAllowed(source, permission) then return true end
-    elseif type(permission) == "table" then
+    end
+
+    if type(permission) == "table" then
```
Manason, line 339: I also think this function is less readable than before. The guard clauses only improve readability when the guarded condition is the edge case and smaller than the "normal" case.

server/functions.lua
```diff
@@ -335,7 +334,9 @@ end
 function QBCore.Functions.HasPermission(source, permission)
     if type(permission) == "string" then
         if IsPlayerAceAllowed(source, permission) then return true end
```
Manason, line 336: nested if can be combined into a single if with an and clause.

Manason (changes requested):
Aside from two of the functions mentioned in the review, I like the guard functions in other places.

server/functions.lua
HebiKotei: I think for this particular case this is very debatable, this is also for extensibility, for example, if you wanted to add another requirement/check you can just do another if statement underneath and then looks far better than a series of elseif.

server/functions.lua
HebiKotei: I was unhappy with this one as well, I'll revert it back to the original until i can find another way to tidy it up.

ChatDisabled:
Someone else besides me that uses guard clauses? Is it possible to learn this power?

PR #10, opened 2022-10-11
Convert Notify Function to Ox-lib

Manason:
## Description

- Changed Notify function to call ox-lib.notify with backwards compatibility. Deprecated in favor of newly created NotifyV2 function.
- Created new NotifyV2 function which allows for making use of more advanced functionality ox-lib offers.
- NotifyV2 does not allow for setting custom css, unlike ox-lib, as notification styling should be consistent within the framework.

## Checklist

- [x] I have personally loaded this code into an updated qbcore project and checked all of its functionality.


client/functions.lua
```diff
@@ -85,28 +85,43 @@ RegisterNUICallback('getNotifyConfig', function(_, cb)
...
+---@field position? NotificationPosition
+---@field type? NotificationType
+---@field icon? string https://fontawesome.com icon name
+---@field iconColor? string css color value for the icon
+
+---Text box popup for player which dissappears after a set time.
+---@param props NotifyProps
+function QBCore.Functions.NotifyV2(props)
+    props.style = nil
+    lib.notify(props)
```
Manason, line 107: Not sure how to prevent lib from being a lint failure. I was looking at qb-weed as an example and didn't see anything being done different.

JericoFX: what if you try to do local lib = lib on top?

D4isDAVID: The linter doen't support ox_lib at the moment, just ignore it.

Molicheu (approved):
These work fine for me, tested on multiple resources using QBCore Notify and had no issues. Saves a lot of time when converting over

Molicheu (changes requested):
image
I'm getting these errors in the console

Manason:
> image
> I'm getting these errors in the console

Renamed the type parameter as this is a keyword already used in lua.

BerkieBb:
Adjust the lint.yaml file to add the library `ox_lib`

BerkieBb:
It still fails, but it won't fail on the commit after this PR has been merged

HebiKotei:
> Adjust the lint.yaml file to add the library `ox_lib`

Considering how long this PR has been up, I have went ahead and just added it to the branch myself. However, it seems to still not work on some lib calls.

TonybynMp4:
what's the point of notifyV2 ? at this point just use ox lib's function 🤔

Manason:
> what's the point of notifyV2 ? at this point just use ox lib's function 🤔

Dependency abstraction has several positive qualities

- We can change easily change NotifyV2 to use something other than ox_lib in the future
- We can control breaking changes. Say ox_lib decides to update and create a breaking change. By having an abstraction layer, we don't have to persist that breaking change through our API, and therefore have control over it rather than giving away control to ox.
- We can restrict ox_lib functionality. NotifyV2 has the opinion that notify CSS should not be different across the qb framework. Whereas ox_lib has this ability, we can have a different opinion using an abstraction layer.

PR #17, opened 2022-11-08
feat(server): check duplicate licenses on player joining

D4isDAVID:
## Description

Needs feedback & testing, maybe it doesn't work and/or maybe there's a better solution.

Fixes QBCore issue qbcore-framework/qb-core#823

## Checklist

- [x] I have personally loaded this code into an updated qbcore project and checked all of its functionality.

### DIFF
--- a/server/events.lua
+++ b/server/events.lua
@@ -15,7 +15,7 @@ AddEventHandler('playerJoining', function()
     local license = QBCore.Functions.GetIdentifier(src, 'license')
     if not license then return end
     if usedLicenses[license] then
-        Wait(0)
+        Wait(0) -- mandatory wait for the drop reason to show up
         DropPlayer(src, Lang:t('error.duplicate_license'))
     else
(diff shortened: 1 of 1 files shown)


Manason (approved):

server/events.lua
```diff
@@ -1,14 +1,31 @@
...
         CancelEvent()
         return
     end
 end)
 
+AddEventHandler('playerJoining', function()
+    if not QBConfig.Server.CheckDuplicateLicense then return end
+    local src = source
+    local license = QBCore.Functions.GetIdentifier(src, 'license')
+    if not license then return end
```
Manason, line 16: If the player doesn't have a license, should they be dropped as well?

D4isDAVID: No, this is already checked when the player connects, in the case that this is enabled in the config.

Manason (approved):
LGTM

PR #24, opened 2022-11-21
fix(notify): fixed config position option

### LINKED ISSUE #23: [SUGGESTION] notify
### The problem

Hi there, So far QBox has been awesome for us, However No matter what position setting I put in the config.lua the notifications seem to default to top right, I was wonder if this is just incomplete? Also an option for button center would be amazing

### Ideal solution

More position options for qb:notify

### Alternative solutions

_No response_

### Additional context

_No response_

Manason:
## Description

fixes #23 
- fixed config position option for notifications to actually modify the notification box position.
- removed other notify config options
- removed/deleted UI files handling notifications as these are now handled by ox_lib

## Checklist


YvanCywan (approved):
LGTM

config.lua
```diff
@@ -32,33 +32,5 @@ QBConfig.Server.Permissions = { 'god', 'admin', 'mod' } -- Add as many groups as
...
-    police = {
-        classes = 'police',
-        icon = 'local_police'
-    },
-    ambulance = {
-        classes = 'ambulance',
-        icon = 'fas fa-ambulance'
-    }
+    position = 'top-right', -- 'top' | 'top-right' | 'top-left' | 'bottom' | 'bottom-right' | 'bottom-left'
 }
```
YvanCywan, line 36: Couldn't this be trimmed further towards something like
```
QBConfig.Notify.Position = 'top-right'
```

Manason: Yeah, it could be. I wasn't sure if I wanted to change the original structure. What do you think?

YvanCywan: Probably better since now you are just outputting an array/table containing one value. Especially since ox_lib will just take a string as an argument.

The only thing to watch out for would be any bits of code still using the old notify system.

Manason: Simplified to just a string.

JoeSzymkowiczFiveM (approved):
Looks good.

PR #26, opened 2022-12-01
(Refactor) Extracted Unique ID Functions

YvanCywan:
## Description

Extracted repeated code to do with id creation (as well as phone number and all that bs) in order to cut down repeated code. Reduced several functions to one as well as providing a better way create new identifiers by modifying the config.

## Checklist

### DIFF
--- a/config.lua
+++ b/config.lua
@@ -17,6 +17,44 @@ QBConfig.Player.ThirstRate = 3.8 -- Rate at which thirst goes down.
 QBConfig.Player.Bloodtypes = {
     "A+", "A-", "B+", "B-", "AB+", "AB-", "O+", "O-",
 }
+QBConfig.Player.IdentifierTypes = {
+    ['citizenid'] = {
+        ValueFunction = function()
+            return tostring(QBCore.Shared.RandomStr(3) .. QBCore.Shared.RandomInt(5)):upper()
+        end,
+        DatabaseColumn = 'citizenid'
+    },
+    ['AccountNumber'] = {
+        ValueFunction = function()
+            return 'US0' .. math.random(1, 9) .. 'QBCore' .. math.random(1111, 9999) .. math.random(1111, 9999) .. math.random(11, 99)
+        end,
+        DatabaseColumn = 'charinfo'
+    },
+    ['PhoneNumber'] = {
+        ValueFunction = function()
+            return math.random(100,999) .. math.random(1000000,9999999)
+        end,
+        DatabaseColumn = 'charinfo'
+    },
+    ['FingerId'] = {
+        ValueFunction = function()
+            return tostring(QBCore.Shared.RandomStr(2) .. QBCore.Shared.RandomInt(3) .. QBCore.Shared.RandomStr(1) .. QBCore.Shared.RandomInt(2) .. QBCore.Shared.RandomStr(3) .. QBCore.Shared.RandomInt(4))
+        end,
+        DatabaseColumn = 'metadata'
+    },
+    ['WalletId'] = {
+        ValueFunction = function()
+            return 'QB-' .. math.random(11111111, 99999999)
+        end,
+        DatabaseColumn = 'metadata'
+    },
+    ['SerialNumber'] = {
+        ValueFunction = function()
+            return math.random(11111111, 99999999)
+        end,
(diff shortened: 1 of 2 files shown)


server/player.lua
```diff
@@ -620,86 +620,46 @@ function QBCore.Player.GetFirstSlotByItem(items, itemName)
...
-    while not UniqueFound do
-        SerialNumber = math.random(11111111, 99999999)
-        local query = '%' .. SerialNumber .. '%'
-        local result = MySQL.prepare.await('SELECT COUNT(*) as count FROM players WHERE metadata LIKE ?', { query })
-        if result == 0 then
-            UniqueFound = true
-        end
-    end
-    return SerialNumber
+    local rand = math.random(11111111, 99999999)
```
BerkieBb, line 663: I don't think math.random()() does something

BerkieBb (changes requested):

server/player.lua
YvanCywan: Yeah, this one in particular will only execute once, changing it to the same style as others. Unless you are saying math.random() doesn't exist... which it does.

Manason (changes requested):
Great refactor!

server/player.lua
```diff
@@ -621,85 +621,50 @@ end
...
-function QBCore.Player.CreateCitizenId()
-    local UniqueFound = false
-    local CitizenId = nil
-    while not UniqueFound do
-        CitizenId = tostring(QBCore.Shared.RandomStr(3) .. QBCore.Shared.RandomInt(5)):upper()
-        local result = MySQL.prepare.await('SELECT COUNT(*) as count FROM players WHERE citizenid = ?', { CitizenId })
-        if result == 0 then
-            UniqueFound = true
+---Generate unique values for the database
+---@param valGen fun(): string | integer
```
Manason, line 625:
```suggestion
---@param valGen fun(): string | number randomly generates a new value
```

Manason (approved):

YvanCywan:
@Manason I've re requested the review since I had a new idea about how to achieve this.

config.lua
```diff
@@ -17,6 +17,44 @@ QBConfig.Player.ThirstRate = 3.8 -- Rate at which thirst goes down.
...
+        end,
+        DatabaseColumn = 'metadata'
+    },
+    ['SerialNumber'] = {
+        ValueFunction = function()
+            return math.random(11111111, 99999999)
+        end,
+        DatabaseColumn = 'metadata'
+    },
+}
```
Manason, line 57: This structure seems worse to me than having the explicit functions as you did previously. Why did you make this change?

YvanCywan: Previously, explicit function's were only used once only, like this we can remove the remainder of the repeating code.
This also allows people to make new identifier types with their own formats by making a config option rather than creating a new explicit function and adding the sql.

Also beyond this, the config also makes it easier for people to change the formats of things like phone numbers etc without having to go digging through some code.

Manason: Moving the shared code to a single function is a good change. The change I'm referring to is moving the generation functions from explicit functions to implied functions in this table via config. This is not getting rid of or extracting any code, but merely changing what was once an explicit function to an entry in a table. Explicit functions are better than implicit functions in a table. The biggest reason is it is easier to trace the callstack of explicit functions than to understand how the table functions are being used.

YvanCywan: Okay, I see your point. In my eyes, I think pushing as much values as possible to the config is never really a bad idea (as long as it remains static, which this is).

Manason: The pattern it sets up though is an API where the user calls one function with a string argument to represent the underlying generation function. I find this layer of abstraction to be more confusing than helpful. I won't revoke my approval over it, but I stand in protest to table based functions like this.

YvanCywan: The table based function is simply so that I can push the formatting function into the config. There are a lot of servers out there that need to change the style of phone numbers (for example) so that they can fit the format of their host country without having to dick around with the function, but yes, this also can be done with functions similarly to before.

Again, I do see your point, and to some degree agree with it. I'd like to wait for some fge

ChatDisabled:
ah yes another Lua fake switch case. Can't wait for people to get confused on this one

PR #27, opened 2022-12-01
(feat) Automatic Issue Assignment on Reply

YvanCywan:
## Description

This allows us to "take" issues by commenting on them. Should just need to respond to a issue with "take" and then you'll be assigned the issue to show that you are working on it. This can also make it easier for people outside the Org to see what's being worked on, and then allow them to even take issues without someone assigning them.

This is potentially a very useful CI change that could help us in the future.

## Checklist

### DIFF
--- a/.github/workflows/assign.yml
+++ b/.github/workflows/assign.yml
@@ -8,11 +8,13 @@ permissions:
 
 jobs:
   issue_assign:
-    issues: write
-    pull-requests: write
-  runs-on: ubuntu-latest
-  steps:
-  - if: github.event.comment.body == 'take'
-    run: |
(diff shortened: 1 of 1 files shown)


Manason:
Aren't issue assignments a native feature of Github? What does this do that can't be done natively?

YvanCywan:
> Aren't issue assignments a native feature of Github? What does this do that can't be done natively?

As we talked about when I messaged you, I will provide a better explanation here for others.

The entire right panel on Issues and PRs are not able to be edited by those outside the GitHub org (rightfully so), but they should still be able to claim issues for their own so that we don't have 2-3+ people working on the same bit of code. After this change, simply responding to an issue with "take" will automatically assign the issue to you, regardless of you perms on the org.

Manason (approved):

PR #29, opened 2023-01-06
tweak(shared): new way to update QBShared

BerkieBb:
## Description

Not tested at all.

## Checklist


client/events.lua
```diff
@@ -220,17 +220,4 @@ RegisterNetEvent('QBCore:Command:ShowMe3D', function(senderId, msg)
...
-    QBCore.Shared[tableName][key] = value
-    TriggerEvent('QBCore:Client:UpdateObject')
-end)
-
-RegisterNetEvent('QBCore:Client:OnSharedUpdateMultiple', function(tableName, values)
-    for key, value in pairs(values) do
-        QBCore.Shared[tableName][key] = value
-    end
-    TriggerEvent('QBCore:Client:UpdateObject')
-end)
```
Manason, line 236: Is this a breaking change, or are these "internal" events? If breaking, add a "!" to PR name. https://www.conventionalcommits.org/en/v1.0.0/

Manason (approved):

client/events.lua
BerkieBb: These are "internal", besides the update object, but it's not really breaking because you won't notice it at all in other scripts and it doesn't break anything by having it removed

BerkieBb:
To be continued...

PR #32, opened 2023-01-31
refactor(client): HasItem: use ox_inventory export

awojnarek:
QBCore.Functions.HasItems just returned back a boolean from qb-inventory. ox_inventory has something very similar, but does not return back a boolean - it returns back a table with k/v pairs with the item and the count that the player has.

This refactor calls the ox_inventory export, and does the logic to determine whether or not it's a table or a single string to check against. This allows you to:

1. Specify a string: `QBCore.Functions.HasItem('drill', 1)`
2. Specify a table with 1 item: `QBCore.Functions.HasItem({'drill'}, 1)`
3. Specify a table with many items: `QBCore.Functions.HasItem({'drill', 'water_bottle'}, 1)`

This allows us to maintain existing functionality for escrowed scripts that use this function.

## Checklist

### DIFF
--- a/client/functions.lua
+++ b/client/functions.lua
@@ -12,6 +12,14 @@ function QBCore.Functions.GetCoords(entity)
     return vector4(coords.x, coords.y, coords.z, GetEntityHeading(entity))
 end
 
+--- QBCore.Functions.HasItem checks if a player has the specified `items` in their inventory
+--- with the specified `amount`.
+---
+--- @param items    The item(s) to check for. Can be a string or a table.
+--- @param amount   The desired quantity of each item
+---
+--- @return boolean Returns true if the player has the specified items in the desired quantity,
+---                 false otherwise
 function QBCore.Functions.HasItem(items, amount)
     local count = exports.ox_inventory:Search('count', items)
 
--- a/client/functions.lua
+++ b/client/functions.lua
@@ -13,7 +13,18 @@ function QBCore.Functions.GetCoords(entity)
 end
 
 function QBCore.Functions.HasItem(items, amount)
-    return exports['qb-inventory']:HasItem(items, amount)
+    local count = exports.ox_inventory:Search('count', items)
+
+    if type(items) == 'table' and type(count) == 'table' then
+        for _, v in pairs(count) do
+            if v < amount then
+                return false
+            end
+        end
+        return true
+    end
+     
+    return count >= amount
 end
 
 -- Utility
--- a/client/functions.lua
+++ b/client/functions.lua
@@ -13,7 +13,8 @@ function QBCore.Functions.GetCoords(entity)
 end
 
 --- QBCore.Functions.HasItem checks if a player has the specified `items` in their inventory
---- with the specified `amount`.
+--- with the specified `amount`. Returns true if the player has at least the amount specified
+--- and not that the player has the exact amount.
 ---
 --- @param items string|string[]    The item(s) to check for. Can be a string or a table and is mandatory.
 --- @param amount? integer          The desired quantity of each item. Acceptable to pass nil, will default to 1.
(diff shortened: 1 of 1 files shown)


BerkieBb (approved):

client/functions.lua
```diff
@@ -13,7 +13,18 @@ function QBCore.Functions.GetCoords(entity)
 end
 
 function QBCore.Functions.HasItem(items, amount)
```
Manason, line 15: Could you write a header for the different overloads of this function? https://github.com/sumneko/lua-language-server/wiki/Annotations#overload

Manason (changes requested):

client/functions.lua
awojnarek: Yessir. I amended the commit - let me know if I need to do anything differently workflow wise once a change has been requested. I see it in the list of commits for this PR.

Manason: Does putting a space after the comment work? I haven't seen that before.

You can put types after the name Ex: @param items string|string[]

Should amount be optional, defaulting to any?

Unclear if items means has to have all the items, or items is if has any of the items.

BerkieBb: The space before param doesn't matter, we just don't do it because it comes without a space on auto-fill

BerkieBb: Also the params now aren't assigned a type, items should be a table and amount should be a number, none optional anymore with this usage but before it was optional

awojnarek: The amount is required. I can make that clearer in the annotations - as well as refactor to basically say if `amount` is nil, then amount = 1. 

Not sure that makes it better per se, but rather enables lazy developers.

awojnarek: Although, reading through the annotation documentation, a param is only required if it does not have a ? before it. So technically everything looks correct.

BerkieBb: Besides the type, it's missing that
`param paramName type`
`param amount number`

awojnarek: > string|string[]

Amended commit.

BerkieBb: I still don't see the types?

awojnarek: Sorry. Should be there now. When I go to amend a commit, it forces me to pull, not sure why because the tip of this branch has not changed. Anyway.

BerkieBb: It's good now yeah

client/functions.lua
```diff
@@ -12,8 +12,27 @@ function QBCore.Functions.GetCoords(entity)
...
+---
+--- @return boolean Returns true if the player has the specified items in the desired quantity,
+---                 false otherwise
 function QBCore.Functions.HasItem(items, amount)
-    return exports['qb-inventory']:HasItem(items, amount)
+    local count = exports.ox_inventory:Search('count', items)
+
+    if type(items) == 'table' and type(count) == 'table' then
+        for _, v in pairs(count) do
+            if v < amount then
```
Manason, line 28: should amount be optional? So that if the caller doesn't specify an amount it defaults to any?

client/functions.lua
Manason: Can you clarify within the documentation that all items are required. I would actually suggest just creating a HasItems function instead of overloading this one, but it's fine.

Manason: Can you also clarify within the documentation that amount is true if player has at least that amount and not player has to have exactly that amount.

awojnarek: @Manason - I have made the requested changes. I did a bit of 'market research' and determined that most people that use this function are checking if it has 1 of the items - so I concluded that it would be best if we allowed the amount to be nil and to default to 1.

Furthermore, the functions name is self-describing - so this **should** be the behavior. The functions name is "HasItem" - so passing along just the item should return a boolean either way.

I have updated the annotations to reflect this - as well as marking amount as optional within annotations.

Manason (approved):

PR #41, opened 2023-02-14
fix(server/storage): make fetch returns optional

Manason:
## Description
- Code previously assumed that bans and players existed within the database if they were being queried. Now there are nil checks and the fetch return types are optional to account for this.

## Checklist

### DIFF
--- a/server/storage.lua
+++ b/server/storage.lua
@@ -49,10 +49,11 @@ end
 ---@field reason string
 
 ---@param request GetBanRequest
----@return BanEntity
+---@return BanEntity?
 function FetchBanEntity(request)
     local column, value = getBanId(request)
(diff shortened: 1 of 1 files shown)


server/storage.lua
```diff
@@ -49,10 +49,11 @@ end
 ---@field reason string
 
 ---@param request GetBanRequest
----@return BanEntity
+---@return BanEntity?
 function FetchBanEntity(request)
     local column, value = getBanId(request)
     local result = MySQL.single.await('SELECT * FROM bans WHERE ' ..column.. ' = ?', { value })
+    if not result then return end
     return {
```
BerkieBb, line 57: You can do
```lua
return result and {
    expire = result.expire,
    reason = result.reason
} or nil
```

BerkieBb (approved):
Indeed I do

D4isDAVID (approved):

PR #44, opened 2023-02-16
Client: Ox Lib conversions and Cleanup

FjamZoo:
## Description
Converted more functions to use ox_lib and generally did some cleanup.

## Checklist

### DIFF
--- a/client/functions.lua
+++ b/client/functions.lua
@@ -256,7 +256,7 @@ function QBCore.Functions.GetPlayersFromCoords(coords, distance)
     local players = lib.getNearbyPlayers(coords, distance or 5, true)
 
     -- This is for backwards compatability as beforehand it only returned the PlayerPed, where Lib returns PlayerPed, PlayerId and PlayerCoords
-    for i = 1, #players, 1 do
+    for i = 1, #players do
         players[i] = players[i].playerPed
     end
 
--- a/client/functions.lua
+++ b/client/functions.lua
@@ -14,7 +14,7 @@ end
 
 --- QBCore.Functions.HasItem checks if a player has the specified `items` in their inventory
 --- with the specified `amount`. Returns true if the player has at least the amount specified
---- and not that the player has the exact amount. If the user passes nil for `amount` then we 
+--- and not that the player has the exact amount. If the user passes nil for `amount` then we
 --- default to 1 - as it's self explainatory within the functions name.
 ---
 --- @param items string|string[]    The item(s) to check for. Can be a string or a table and is mandatory.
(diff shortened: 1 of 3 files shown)


client/functions.lua
```diff
@@ -250,61 +244,40 @@ function QBCore.Functions.IsWearingGloves()
...
 
 function QBCore.Functions.GetClosestPlayer(coords)
-    local ped = PlayerPedId()
-    if coords then
-        coords = type(coords) == 'table' and vec3(coords.x, coords.y, coords.z) or coords
-    else
-        coords = GetEntityCoords(ped)
-    end
-
+    coords = type(coords) == 'table' and vec3(coords.x, coords.y, coords.z) or coords or GetEntityCoords(cache.ped)
```
Manason, line 247: This line is repeated several times. Consider extracting to a local function.

client/functions.lua
```diff
@@ -250,61 +244,40 @@ function QBCore.Functions.IsWearingGloves()
...
-    local players = GetActivePlayers()
-    local ped = PlayerPedId()
-    if coords then
-        coords = type(coords) == 'table' and vec3(coords.x, coords.y, coords.z) or coords
-    else
-        coords = GetEntityCoords(ped)
-    end
-    distance = distance or 5
+    coords = type(coords) == 'table' and vec3(coords.x, coords.y, coords.z) or coords or GetEntityCoords(cache.ped)
+    local Players = lib.getNearbyPlayers(coords, distance or 5, true)
```
Manason, line 256:
```suggestion
    local players = lib.getNearbyPlayers(coords, distance or 5, true)
```

local functions/var names should be camelCase

client/functions.lua
```diff
@@ -250,61 +244,40 @@ function QBCore.Functions.IsWearingGloves()
...
-    for _, player in pairs(players) do
-        local target = GetPlayerPed(player)
-        local targetCoords = GetEntityCoords(target)
-        local targetdistance = #(targetCoords - coords)
-        if targetdistance <= distance then
-            closePlayers[#closePlayers + 1] = player
-        end
+    for _, v in pairs(Players) do
+        closePlayers[#closePlayers+1] = v.playerPed
     end
```
Manason, line 262: Consider using a numeric for loop here. https://manason.github.io/effective-fivem-lua/tables/#use-numeric-for-loops-when-iterating-over-an-array

Manason (approved):

client/functions.lua
FjamZoo: I probably wont make this into a local function would be kinda wierd, the only reason why i reused the code is because it just replaced the 5-6 long lines of wierdness that was going on last time.

Manason: Not sure why it would be weird to extract duplicate code into a function, but up to you.

PR #48, opened 2023-03-08
Fix licensekey error

papabendi:
## Description

Fixes an issue where it sees new characters with a null license key. PlayerData.license is null by default.

## Checklist

- [X ] I have personally loaded this code into an updated Qbox project and checked all of its functionality.
- [ X] My code fits the style guidelines.
- [ X] My PR fits the contribution guidelines.

### DIFF
--- a/server/player.lua
+++ b/server/player.lua
@@ -42,7 +42,7 @@ function QBCore.Player.CheckPlayerData(source, PlayerData)
     local Offline = true
     if source then
         PlayerData.source = source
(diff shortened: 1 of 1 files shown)


BerkieBb:
The license error exists when a player doesn't have a license2, this is only the case when their FiveM hasn't generated it, in that case you fix it by reinstalling FiveM, because every player has to have a license2. What license2 is, is the 'static' license, it doesn't change, whilst the normal license can change so that's why we use it and not the normal license. With the code you provided, it will always use the normal license since no one doesn't have it and that leads to problems which we had before, a players characters not being loaded because their license is different.

PR #51, opened 2023-03-18
Update functions.lua

litsou:
It will return the correct value for backward compatibility

## Checklist


BerkieBb:
You can adjust the comment to your changes as well in the code

BerkieBb:
And change `playerPed` to an underscore

ChatDisabled:
Could I get some context behind this change though? Still not sure what compatibility this is supposed to accommodate

BerkieBb:
> Could I get some context behind this change though? Still not sure what compatibility this is supposed to accommodate

This is supposed to fix the wrong return variable of this function that was supposedly for backwards compatibility as the GetClosestPlayer function should return playerId, not playerPedId which it does now.

PR #63, opened 2023-05-11
Database refactor

TonybynMp4:
## Description

Gotta start somewhere, please contribute to this lol

## Checklist

### DIFF
--- a/config.lua
+++ b/config.lua
@@ -28,11 +28,6 @@ QBConfig.Player.Bloodtypes = {
 ---@alias UniqueIdType 'citizenid' | 'AccountNumber' | 'PhoneNumber' | 'FingerId' | 'WalletId' | 'SerialNumber'
 ---@type table<UniqueIdType, {valueFunction: function}>
 QBConfig.Player.IdentifierTypes = {
-    ['citizenid'] = {
-        valueFunction = function()
-            return MySQL.scalar.await('SELECT citizenid FROM players ORDER BY citizenid DESC LIMIT 1') + 1
-        end,
-    },
     ['AccountNumber'] = {
         valueFunction = function()
             return 'US0' .. math.random(1, 9) .. 'QBCore' .. math.random(1111, 9999) .. math.random(1111, 9999) .. math.random(11, 99)
--- a/config.lua
+++ b/config.lua
@@ -30,7 +30,7 @@ QBConfig.Player.Bloodtypes = {
 QBConfig.Player.IdentifierTypes = {
     ['citizenid'] = {
         valueFunction = function()
-            return tostring(QBCore.Shared.RandomStr(3) .. QBCore.Shared.RandomInt(5)):upper()
+            return MySQL.scalar.await('SELECT citizenid FROM players ORDER BY citizenid DESC LIMIT 1') + 1
         end,
     },
     ['AccountNumber'] = {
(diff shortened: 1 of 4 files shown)


config.lua
```diff
@@ -30,7 +30,7 @@ QBConfig.Player.Bloodtypes = {
 QBConfig.Player.IdentifierTypes = {
     ['citizenid'] = {
         valueFunction = function()
-            return tostring(QBCore.Shared.RandomStr(3) .. QBCore.Shared.RandomInt(5)):upper()
+            return MySQL.scalar.await('SELECT citizenid FROM players ORDER BY citizenid DESC LIMIT 1') + 1
```
Manason, line 33: Is a citizen entry of IdentifierTypes even needed? This is supposed to be for id generation, but if ids are auto_incremented, then I think this can be deleted.

qbcore.sql
```diff
@@ -12,32 +9,16 @@ CREATE TABLE IF NOT EXISTS `players` (
...
-  `discord` varchar(50) DEFAULT NULL,
-  `ip` varchar(50) DEFAULT NULL,
-  `reason` text DEFAULT NULL,
-  `expire` int(11) DEFAULT NULL,
-  `bannedby` varchar(255) NOT NULL DEFAULT 'LeBanhammer',
-  PRIMARY KEY (`id`),
-  KEY `license` (`license`),
-  KEY `discord` (`discord`),
-  KEY `ip` (`ip`)
-) ENGINE=InnoDB AUTO_INCREMENT=1;
```
Manason, line 33: what happened to bans?

TonybynMp4: i guess we really shouldn't manage that ourselves and use txadmin's bans instead cauz txadmin actually maintains that system

qbcore.sql
```diff
@@ -12,32 +9,16 @@ CREATE TABLE IF NOT EXISTS `players` (
   `inventory` longtext DEFAULT NULL,
   `last_updated` timestamp NOT NULL DEFAULT current_timestamp() ON UPDATE current_timestamp(),
   PRIMARY KEY (`citizenid`),
-  KEY `id` (`id`),
+  FOREIGN KEY (`userid`) REFERENCES `users` (`userid`) ON DELETE CASCADE ON UPDATE CASCADE,
```
Manason, line 12: Interesting discussion was had last night regarding updating cached values when deletes/updates are cascaded. While the FK constraint is great, we need to have a discussion on whether we want to cascade deletes/updates on the DB and figure out an alternate cache solution, or otherwise keep the cascades within lua code so that the cache can be easily updated.

qbcore.sql
FjamZoo: Yeah bans should not be handled in the core cuz FUCK me that was a shit system, hopefully TXAdmin intergrates a ban so we can run it through txadmin completely

TonybynMp4: > Yeah bans should not be handled in the core cuz FUCK me that was a shit system, hopefully TXAdmin intergrates a ban so we can run it through txadmin completely

shouldn't be hard to pr exports for their moderation tools 🤔

FjamZoo: > shouldn't be hard to pr exports for their moderation tools 🤔

The hard part is always getting the PR approved

benzon:
After trying out the changes in an environment this is not going to work.

Removing citizenid and not generating it ad the start will cause a heap of problems, since so many scripts relies on it, not sure what the appropriate solution here would be, since the current PR’ed solution seems like it’s not going to fly.

PR #71, opened 2023-05-20
feat(server)!: Vehicle Spawning Overhaul

Manason:
## Description

- deprecating QB callbacks for creating/spawning vehicle in favor of new ox_lib callbacks
- deprecating vehicle creation functions/callbacks on the server which use CreateVehicle native rather than CreateVehicleServerSetter. Existing functions/callbacks are now wrappers.
- BREAKING CHANGE: 'QBCore:Server:CreateVehicle' server callback will now return nil rather than 0 if the server artifact is below 5904. This brings the behavior in-line with the function version.
- server events/functions which spawn vehicle refactored to call the server to spawn the vehicle
- BREAKING CHANGE: QBCore.Functions.CreateVehicle server function now returning the netId instead of the entityId. I'm too sure about whether this is a good change or not, but it seemed odd that the original callback returned the netId and the function returned the entityId, so I picked one and went with it.

Unrelated: I don't understand why the createVehicle function is spawning and deleting a temp vehicle.
Unrelated: Seems like the callback for spawning vehicles should live right next to the server function for spawning vehicles rather than have it in the 'events' file.

## Checklist

### DIFF
--- a/client/events.lua
+++ b/client/events.lua
@@ -1,3 +1,5 @@
+local utils = require 'client.utils'
+
 -- Player load and unload handling
 RegisterNetEvent('QBCore:Client:OnPlayerLoaded', function()
     ShutdownLoadingScreenNui()
@@ -129,22 +131,22 @@ end)
 
 -- Vehicle Commands
 
----internal event to core. Do not invoke.
----@param vehicle number
-RegisterNetEvent('QBCore:Command:SpawnVehicle', function(netId)
+utils.entityStateHandler('initVehicle', function(entity, _, value)
+    if not value or NetworkGetEntityOwner(entity) ~= cache.playerId then return end
     if cache.vehicle then
         DeleteVehicle(cache.vehicle)
     end
 
-    local vehicle
-    repeat
-        vehicle = NetToVeh(netId)
-        Wait(0)
-    until DoesEntityExist(vehicle)
+    local veh = entity
 
-    SetVehicleFuelLevel(vehicle, 100.0)
-    SetVehicleDirtLevel(vehicle, 0.0)
-    TriggerEvent("vehiclekeys:client:SetOwner", QBCore.Functions.GetPlate(vehicle))
+    SetVehicleHasBeenOwnedByPlayer(veh, true)
+    SetNetworkIdCanMigrate(netId, true)
+    SetVehicleNeedsToBeHotwired(veh, false)
+    SetVehRadioStation(veh, 'OFF')
+    SetVehicleFuelLevel(veh, 100.0)
+    SetVehicleDirtLevel(veh, 0.0)
+    TriggerEvent("vehiclekeys:client:SetOwner", QBCore.Functions.GetPlate(veh))
+    Entity(entity).state:set('initVehicle', nil, true)
 end)
 
 RegisterNetEvent('QBCore:Command:DeleteVehicle', function()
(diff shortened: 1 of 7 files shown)


client/events.lua
```diff
@@ -129,26 +131,22 @@ end)
...
+    local veh = entity
+
+    SetVehicleHasBeenOwnedByPlayer(veh, true)
+    SetNetworkIdCanMigrate(netId, true)
+    SetVehicleNeedsToBeHotwired(veh, false)
+    SetVehRadioStation(veh, 'OFF')
+    SetVehicleFuelLevel(veh, 100.0)
+    SetVehicleDirtLevel(veh, 0.0)
+    TriggerEvent("vehiclekeys:client:SetOwner", QBCore.Functions.GetPlate(veh))
+    Entity(entity).state:set('initVehicle', nil, true)
```
FjamZoo, line 149:
```suggestion
    SetVehicleNeedsToBeHotwired(entity, false)
    SetVehRadioStation(entity, 'OFF')
    SetVehicleFuelLevel(entity, 100.0)
    SetVehicleDirtLevel(entity, 0.0)
    TriggerEvent('vehiclekeys:client:SetOwner', QBCore.Functions.GetPlate(veh))
    Entity(entity).state:set('initVehicle', nil, true)
```

Changed here as we don't need veh localized, we also can remove some natives as this is server side created vehicles.

client/events.lua
```diff
@@ -129,26 +131,22 @@ end)
...
-RegisterNetEvent('QBCore:Command:SpawnVehicle', function(vehName)
-    local hash = joaat(vehName)
-    if not IsModelInCdimage(hash) then return end
-    RequestModel(hash)
-    while not HasModelLoaded(hash) do
-        Wait(0)
-    end
-
+utils.entityStateHandler('initVehicle', function(entity, _, value)
+    if not value or NetworkGetEntityOwner(entity) ~= cache.playerId then return end
```
FjamZoo, line 135:
```suggestion
    if not value then return end
    
    for i = -1, 0 do
        local ped = GetPedInVehicleSeat(entity, i)

        if ped ~= cache.ped and ped > 0 and NetworkGetEntityOwner(ped) == cache.playerId then
            DeleteEntity(ped)
        end
    end
    
    if NetworkGetEntityOwner(entity) ~= cache.playerId then return end
```

We need to add this back because sometimes with networked server side vehicles ped spawns inside the vehicle BECAUSE some dumb cfx fuckery

server/functions.lua
```diff
@@ -245,7 +238,9 @@ function QBCore.Functions.CreateVehicle(source, model, coords, warp)
     local veh = CreateVehicleServerSetter(model, vehicleType, coords.x, coords.y, coords.z, coords.w)
     while not DoesEntityExist(veh) do Wait(0) end
     if warp then TaskWarpPedIntoVehicle(GetPlayerPed(source), veh, -1) end
-    return veh
+    local state = Entity(veh).state
+    state:set('initVehicle', true, true)
```
FjamZoo, line 242:
```suggestion
    Entity(veh).state:set('initVehicle', true, true)
```

We do not need state localized because we do not index it multiple times

FjamZoo (changes requested):

FjamZoo (approved):

FjamZoo (approved):

PR #78, opened 2023-06-10
Remove callbacks from core

BerkieBb:
## Description

Here it is...

The long awaited PR to remove...

CALLBACKS

*insert reveal soundtrack*

## Checklist


Manason:
We should probably hold off on this for several months...

benzon:
> We should probably hold off on this for several months...

Would agree with you on that, since it would be ruing a lot of backwards compatibility with qb-core in general.

Takes time for developers to migrate scripts to use ox_lib based callbacks.

PR #80, opened 2023-06-13
Utils import

BerkieBb:
## Description

Adds an import folder to add imports for other resources, this PR only add a utils.lua for utility functions.
Aside from adding that I also cleaned up the annotations

## Checklist


client/functions.lua
```diff
@@ -10,23 +10,22 @@ function QBCore.Functions.GetPlayerData(cb)
     cb(QBCore.PlayerData)
 end
 
+---@deprecated use GetCoordsFromEntity from imports/utils.lua
 ---@param entity number
 ---@return vector4
 function QBCore.Functions.GetCoords(entity)
     local coords = GetEntityCoords(entity)
     return vector4(coords.x, coords.y, coords.z, GetEntityHeading(entity))
```
Manason, line 18: Possible to call the utils function here to remove the implementation code?

Manason (changes requested):

imports/utils.lua
```diff
@@ -0,0 +1,1007 @@
+local isServer = IsDuplicityVersion()
```
Manason, line 1: Consider putting a comment at the top of this file that summarizes how it's used, and what's in it.

BerkieBb: I don't think that's a good idea as the modules that I have ready will change the system up a bit, so you would import a different file that handles the importing of these files.

PR #83, opened 2023-06-17
feat(import): resource name

BerkieBb:
## Description

Adds the ability to specify a resource name with the module loading

## Checklist

### DIFF
--- a/import.lua
+++ b/import.lua
@@ -22,6 +22,38 @@ local function checkSide(value)
     return value == 'client' and not isServer or value == 'server' and isServer or value == 'shared'
 end
 
+local function checkModule(module, resourceName)
+    if not module then return end
+
+    local file = module
+    local split = {string.strsplit(':', module)}
+    local doLoad = true
+    if #split > 1 then
+        local side = getSide(split[1])
+        local isSide = side ~= 'none'
+        local isResource = resources[split[1]]
+        if isResource then
+            resourceName = split[1]
+            side = getSide(split[2])
+            isSide = side ~= 'none'
+        end
+
+        side = not isSide and 'shared' or side
+
+        doLoad = checkSide(side)
+        file = isResource and isSide and split[3] or not isResource and isSide and split[2] or split[1]
+    end
+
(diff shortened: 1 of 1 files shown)


import.lua
```diff
@@ -5,22 +5,51 @@
...
+        'resourceName:client:module_name' -- The resource loading also accepts a side to load on
     }
 ]]
 
-local resourceName = 'qbx-core'
+local resources = {}
 local isServer = IsDuplicityVersion()
 
+local function getSide(value)
+    return value == 'client' and 'client' or value == 'server' and 'server' or value == 'shared' and 'shared' or 'none'
```
Manason, line 18:
```suggestion
    if value == 'client' or value == 'server' or value == 'shared' then return value
    else return 'none' end
```

nit; I find this way to be a bit more readable.

import.lua
```diff
@@ -5,22 +5,51 @@
...
+end
+
+for i = 0, GetNumResources() - 1 do
+    local resource = GetResourceByFindIndex(i)
+    if GetResourceState(resource) == 'started' then
+        resources[resource] = true
+    end
+end
+
 for i = 0, GetNumResourceMetadata(resourceName, 'module') - 1 do
```
Manason, line 32: Is this function only called once to get the number of resource metadata? Or is it called on every iteration?

import.lua
```diff
@@ -5,22 +5,51 @@
...
+    local resourceName = 'qbx-core'
     local module = GetResourceMetadata(resourceName, 'module', i)
     if module then
```
Manason, line 35: Consider extracting a function and then inverting to use more guard clauses.

Manason (approved):

import.lua
BerkieBb: GetNumResourceMetadata will only be called once to get the number to reach until it needs to end

import.lua
BerkieBb: It seems readable enough for me as it is, pretty straight forward

Manason (approved):

PR #87, opened 2023-06-24
staff permissions

TonybynMp4:
## Description

Adds permissions when player connects, fixes issue players seems to have currently where setting principals to identifiers doesn't work (but player.[source] works somehow)

+it better than adding lines in server.cfg (i guess?)

The Way i did it definitely isn't the best but it might be easier for config devs (i think)
If ya want i could just use only one identifier type and put the identifier as key, would remove loops.
Could also have a table for each identifier type i guess?
also might need to be addprincipals instead of ace? i don't really understand the whole principals/ace stuff (like i get aces but then what the hell are principals for?????)

## Checklist

### DIFF
--- a/fxmanifest.lua
+++ b/fxmanifest.lua
@@ -32,6 +32,7 @@ server_scripts {
     'server/debug.lua',
     'server/loops.lua',
     'server/storage.lua',
+    'server/staff.lua'
 }
 
 modules {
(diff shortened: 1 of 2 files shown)


BerkieBb:
We don't need this, you set the ace in the server.cfg already

TonybynMp4:
> We don't need this, you set the ace in the server.cfg already

It brokie tho.

TonybynMp4:
ok me dumb

PR #102, opened 2023-07-22
fix(modules/utils): ChangeVehicleExtra stack overflow

buddizer:
## Description

- Fixes checking for extras
- Resets extras if found
- Applies extras properly

## Checklist

### DIFF
--- a/modules/utils.lua
+++ b/modules/utils.lua
@@ -895,11 +895,14 @@ else
     ---@param enable boolean
     function ChangeVehicleExtra(vehicle, extra, enable)
         if not DoesExtraExist(vehicle, extra) then return end
-
-        SetVehicleExtra(vehicle, extra, not enable)
         local isExtraOn = IsVehicleExtraTurnedOn(vehicle, extra)
+
         if enable ~= isExtraOn then
-            ChangeVehicleExtra(vehicle, extra, enable)
+            SetVehicleExtra(vehicle, extra, not enable)
+        end    for i = 1, 20 do
+            if DoesExtraExist(vehicle, i) then
+                SetVehicleExtra(vehicle, i, false)
+            end
         end
     end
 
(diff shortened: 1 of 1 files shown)


modules/utils.lua
```diff
@@ -910,7 +914,7 @@ else
         -- Clear Extras
         for i = 1, 20 do
             if DoesExtraExist(vehicle, i) then
-                SetVehicleExtra(vehicle, i, true)
+                SetVehicleExtra(vehicle, i, false)
```
Manason, line 917: consider extracting a local function clearVehicleExtras to reduce duplicate code

Manason (approved):

modules/utils.lua
buddizer: How about something like this?

`    ---Clear all vehicle extras
    ---@param vehicle integer
    function ClearAllVehicleExtras(vehicle)
        for i = 1, 20 do
            if DoesExtraExist(vehicle, i) then
                SetVehicleExtra(vehicle, i, false)
            end
        end
    end

    ---Set the status of an extra on the vehicle
    ---@param vehicle integer
    ---@param extra integer
    ---@param enable boolean
    function ChangeVehicleExtra(vehicle, extra, enable)
        if not DoesExtraExist(vehicle, extra) then return end
        local isExtraOn = IsVehicleExtraTurnedOn(vehicle, extra)

        if enable ~= isExtraOn then
            SetVehicleExtra(vehicle, extra, not enable)
        end
    end

    ---Set the vehicle extras of a vehicle according to a table
    ---@param vehicle integer
    ---@param extras table<integer, boolean>
    function SetVehicleExtras(vehicle, extras)
        ClearAllVehicleExtras(vehicle)
        
        for id, enabled in pairs(extras) do
            ChangeVehicleExtra(vehicle, tonumber(id) --[[@as integer]], enabled)
        end
    end`

PR #106, opened 2023-08-03
refactor(callback): change to ox_lib callbacks

ddoucethollett:
## Description
changed the callbacks in the core to call the equivalent callbacks in ox_lib. as well as removing calling client callbacks and server callbacks from their own side (just weird and anyone using it isn't using code right).

changed the empty functions to use ox_lib/init.lua noop function to reduce memory usage slightly by not defining additional functions.

## Checklist

### DIFF
--- a/client/events.lua
+++ b/client/events.lua
@@ -173,25 +173,6 @@ RegisterNetEvent('QBCore:Notify', function(text, notifyType, duration, subTitle,
     QBCore.Functions.Notify(text, notifyType, duration, subTitle, notifyPosition, notifyStyle, notifyIcon, notifyIconColor)
 end)
 
--- Callback Events --
-
--- Client Callback
----@deprecated call a function instead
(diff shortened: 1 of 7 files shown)


Manason (approved):

l3oncoder:
I did test these changes, and notice breaks stuff in relation to cb(), Do you know a solution for this 🤔

Manason:
> I did test these changes, and notice breaks stuff in relation to cb(), Do you know a solution for this 🤔

Reviewing the code again, we need to do some conversion to call the ox_lib callbacks, to convert the parameters of qb cb to ox_lib cb

ddoucethollett:
Totally right. will update.

ddoucethollett:
Although the args from cb should be passed, because the cb is being passed directly. and the args should be the same, `source, ...` source always being the first argument, at least with server callbacks and then everything else. Will test more and update though.

PR #118, opened 2023-09-07
Chore/add bring command

kylemcshea:
added a `/bring [id]` command to be utilized by admins.

ensures the following cases are handled.
1. id is non-numeric. in this case will give a notification asking for valid input.
2. target is not online. in this case will notify player the target is not online.
3. player tries to /bring themselves. in this case it tells player cannot teleport to themselves.

I can work on implementing localization if this is a pr that you guys would like to be added to the project. I am not sure if this functionality is already handled in separate command.

### DIFF
--- a/server/commands.lua
+++ b/server/commands.lua
@@ -65,6 +65,33 @@ lib.addCommand('tp', {
     end
 end)
 
+lib.addCommand('bring', {
+    help = 'bring player to you by id',
+    params = {
+        { name = 'id', help = 'id of player to bring', optional = false },
+    },
+    restricted = "group.admin"
+}, function(source, args)
+    if not args['id'] then
+        TriggerClientEvent('QBCore:Notify', source, 'We need an id to teleport to.', 'error')
+        return
+    end
+
+    local targetId = tonumber(args['id'])
+    local target = GetPlayerPed(targetId)
+    if source == targetId then
+        return TriggerClientEvent('QBCore:Notify', source, 'You cannot teleport to yourself.', 'error')
+    end
+
+    if target ~= 0 then
+        local playerPed = GetPlayerPed(source)
+        local coords = GetEntityCoords(playerPed)
+        TriggerClientEvent('QBCore:Command:TeleportToPlayer', targetId, coords)
+    else
+        TriggerClientEvent('QBCore:Notify', source, Lang:t('error.not_online'), 'error')
+    end
+end)
+
 
 lib.addCommand('tpm', {
     help = Lang:t("command.tpm.help"),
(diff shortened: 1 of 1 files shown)


server/commands.lua
```diff
@@ -65,6 +65,35 @@ lib.addCommand('tp', {
...
 end)
 
+lib.addCommand('bring', {
+    help = 'bring player to you',
+    params = {
+        { name = 'id', help = 'id to bring', optional = false },
+    },
+    restricted = "group.admin"
+}, function(source, args)
+    if not args['id'] or not tonumber(args['id']) then
```
Manason, line 75:
```suggestion
    if not args.id or not tonumber(args.id) then
```

Manason: Change in other places please

server/commands.lua
```diff
@@ -65,6 +65,35 @@ lib.addCommand('tp', {
...
+        TriggerClientEvent('QBCore:Notify', source, Lang:t('error.not_online'), 'error')
+        return
+    elseif source == targetId then
+        TriggerClientEvent('QBCore:Notify', source, 'You cannot teleport to yourself.', 'error')
+        return
+    end
+
+    local playerPed = GetPlayerPed(source)
+    local coords = GetEntityCoords(playerPed)
+    TriggerClientEvent('QBCore:Command:TeleportToPlayer', targetId, coords)
```
Manason, line 93: Would it be better to set the target coords directly here instead of calling this event? We can just set the coordinate from the server side.

Manason (changes requested):

Manason:
Also wondering if the bring command belongs in https://github.com/Qbox-project/qbx-adminmenu/blob/main/server/commands.lua. It seems weird we have two different places where admin commands live.

server/commands.lua
kylemcshea: Yeah i agree I'll make that change

server/commands.lua
kylemcshea: Will do

kylemcshea:
I agree admin commands living in two separate spots is odd.

I feel we should move all these commands in qbx-core to qbx-admin because it would make sense where code lives wise.

With that being said I think if we keep code in two separate spots we should keep `bring [id]` here because more tp commands live on `qbx-core` at the moment

BerkieBb:
> I agree admin commands living in two separate spots is odd.
> 
> I feel we should move all these commands in qbx-core to qbx-admin because it would make sense where code lives wise.
> 
> With that being said I think if we keep code in two separate spots we should keep `bring [id]` here because more tp commands live on `qbx-core` at the moment

These commands are better in qbx-adminmenu yeah

Manason:
Re-open this PR on qbx-adminmenu please.

PR #121, opened 2023-09-09
feat: release and versioning automation

Manason:
## Description

- shamelessly stolen from ox
- can't be done centrally as the bundle steps reference files specific to this repository


BerkieBb:
This can go into the github-docs repo, only thing you need to do then is include the repo name in the zip file instead of hardcoding

Manason:
> This can go into the github-docs repo, only thing you need to do then is include the repo name in the zip file instead of hardcoding

The files would need to be passed then as well. Also if there are web files to build there's a bunch of additional steps to add.

TheiLLeniumStudios:
As mentioned in discord, I'll setup a custom workflow that can be distributed across all repos and it will take care of the build steps if they are needed. So this can be closed

Manason:
Will use this for now until Illenium finishes the global workflow

PR #122, opened 2023-09-10
Convert string through ox_lib for multichar

BerkieBb:
## Description

Implements https://github.com/overextended/ox_lib/pull/411

## Checklist


StrixLuca:
Nice bro

client/character.lua
```diff
@@ -168,14 +168,14 @@ end
...
     for i = 5, 1, -1 do
         local str = dialog[input]
         local pattern = createPattern(i)
 
-        matched = not string.match(str, '^%s') -- Don't match if there is a trailing whitespace at the beginning
-        matched = not string.match(str, '%s$') -- Don't match if there is a trailing whitespace at the end
-        matched = string.match(str, pattern)
+        -- ^%s matches if there is a trailing whitespace at the beginning
+        -- %s$ matches if there is a trailing whitespace at the end
+        matched = not string.match(str, '^%s') and not string.match(str, '%s$') and string.match(str, pattern)
```
Manason, line 182: This function could use a description next to the params, or even an overall description. It looks like the purpose is to return true if all conditions match for all dialog inputs, and false if not? I'm not confident about that but if this is the intention then this would work better with local matched = true and then various conditions cause it to return false immediately.

Manason (changes requested):

client/character.lua
BerkieBb: It's purpose is basically to check the strings but go through a max of 5 words counting from 5 down, so that's why it doesn't return immediately

BerkieBb: Otherwise it would only allow 5 words or more to be in the input fields

Manason: dialog is marked any[], is it really string[]? Seems like it would break if the input wasn't a string.

I'm still unclear why it doesn't return immediately? What is the overall purpose of this method? To check for profanity?

BerkieBb: It's to make sure that the input fields don't have more than 5 words in them, don't include anything but letters, every word starts with a capital, has no trailing whitespace and check for profanity

BerkieBb: The way I check for the amount of words makes it not be able to return immediately

BerkieBb: And dialog is any[] because I couldn't find how to make a fixed array type as it's mixed at the moment, but I can change it to string once the ox pr is merged actually

Manason: I'm still not understanding why you can't return immediately? Why does it count down from 5 rather than count up?

I'm not sure the nots in this case are helping the readability. Maybe it would be easier to understand using if statements?

BerkieBb: It's how I made the check for matching, the matching pattern goes something like this for 3 words `%u%l*%s%u%l*%s%u%l*`. So it only matches when there are 3 words exactly that have an uppercase letter at the beginning of each word and a whitespace in between them, example `This Is True`, wrong example, `This is True`. So I have to check by counting down because if I'm counting up, it won't match the first time if there is more than one word and then if there isn't 5 words, it will always return false when it shouldn't.

Manason: It seems like we're trying to do too much with one pattern. Maybe try breaking up the different validations and a better structure may reveal itself.

Manason (approved):
Nice refactor

Manason (approved):

PR #123, opened 2023-09-11
Update en.lua

alberttheprince:
## Description

Changed Display of Gender to Sex in English locales.

### DIFF
--- a/locale/en.lua
+++ b/locale/en.lua
@@ -54,7 +54,7 @@ local Translations = {
         first_name = 'First Name',
         last_name = 'Last Name',
         nationality = 'Nationality',
-        gender = 'Gender',
+        gender = 'Sex',
         birth_date = 'Birth Date'
     },
     command = {


BerkieBb:
You could've gone a little further and updated the code and name of the translation as well

Manason (approved):

alberttheprince:
> You could've gone a little further and updated the code and name of the translation as well

I wasn't sure if gender was generally used by many different resources, so I only changed it in locales in case other resources rely on code with the term gender in it. Sorry if this isn't good enough, I can look into replacing all of them!

BerkieBb:
> > You could've gone a little further and updated the code and name of the translation as well
> 
> I wasn't sure if gender was generally used by many different resources, so I only changed it in locales in case other resources rely on code with the term gender in it. Sorry if this isn't good enough, I can look into replacing all of them!

It's only used in client/character.lua once, it's been added only yesterday

alberttheprince:
> > > You could've gone a little further and updated the code and name of the translation as well
> > 
> > 
> > I wasn't sure if gender was generally used by many different resources, so I only changed it in locales in case other resources rely on code with the term gender in it. Sorry if this isn't good enough, I can look into replacing all of them!
> 
> It's only used in client/character.lua once, it's been added only yesterday

I meant more like outside resources, unless this is handled here differently than multichar handled it

BerkieBb:
> > > > You could've gone a little further and updated the code and name of the translation as well
> > > 
> > > 
> > > I wasn't sure if gender was generally used by many different resources, so I only changed it in locales in case other resources rely on code with the term gender in it. Sorry if this isn't good enough, I can look into replacing all of them!
> > 
> > 
> > It's only used in client/character.lua once, it's been added only yesterday
> 
> I meant more like outside resources, unless this is handled here differently than multichar handled it

I don't get it, it's just a string...

alberttheprince:
OK Mr. BerkieB, I may be too dense to understand, but it seems like other than the locale these other references are referring to charinfo.gender, which (I think?) shouldn't be changed. 

Apologies for creating any confusion, I just wanted to change it so end users/players would have a more accurate description of what they were choosing in that category. 

I assume

```
                Gender = character.charinfo.gender == 0 and Lang:t('info.char_male') 
                ```
                Shouldn't change to Sex = or sex =

PR #128, opened 2023-09-13
support for differently named identifier columns

hakanesnn:
## Checklist

### DIFF
--- a/config.lua
+++ b/config.lua
@@ -109,4 +109,21 @@ QBConfig.Server.Discord = "" -- Discord invite link
 QBConfig.Server.CheckDuplicateLicense = true -- Check for duplicate rockstar license on join
 QBConfig.Server.Permissions = { 'god', 'admin', 'mod' } -- Add as many groups as you want here after creating them in your server.cfg
 
+QBConfig.Storage = {}
+QBConfig.Storage.PlayerDataTables = {
+    ['players'] = 'citizenid',
+    ['apartments'] = 'citizenid',
+    ['bank_accounts'] = 'citizenid',
+    ['crypto_transactions'] = 'citizenid',
+    ['phone_invoices'] = 'citizenid',
+    ['phone_messages'] = 'citizenid',
+    ['playerskins'] = 'citizenid',
+    ['player_contacts'] = 'citizenid',
+    ['player_houses'] = 'citizenid',
+    ['player_mails'] = 'citizenid',
+    ['player_outfits'] = 'citizenid',
+    ['player_vehicles'] = 'citizenid',
+    -- ['bank_accounts_new'] = 'creator'
+}
+
 QBConfig.NotifyPosition = 'top-right' -- 'top' | 'top-right' | 'top-left' | 'bottom' | 'bottom-right' | 'bottom-left'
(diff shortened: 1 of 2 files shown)


Manason:
I'm not sure the database schema should be something open to be dynamically configured. Why are you desiring to rename the id column?

Manason:
> When I install with qbox txadminrecipe, renewed-banking is installed, renewed-banking creates the created accounts to the db with the creator column, not the citizenid column, I think that when the character is deleted, the accounts belonging to the character should also be deleted. this is just an example of renewed-banking, other script creators may not save the data to a column named citizenid when saving to the db, I needed something like this because the characters are now managed from within qbx-core.

Could we solve this problem by triggering an event when player is deleted and then you could clean up non qbx tables by consuming that event?

server/storage.lua
```diff
@@ -225,28 +225,12 @@ end
...
-        'playerskins',
-        'player_contacts',
-        'player_houses',
-        'player_mails',
-        'player_outfits',
-        'player_vehicles',
-    }
-
-    local query = "DELETE FROM %s WHERE citizenid = ?"
+    local query = "DELETE FROM %s WHERE %s = ?"
```
FjamZoo, line 228:
```suggestion
    local query = 'DELETE FROM %s WHERE %s = ?'
```

pretty please thank you

FjamZoo:
> Could we solve this problem by triggering an event when player is deleted and then you could clean up non qbx tables by consuming that event?

or imagine this we could use something thats called foreign keys 🤯

config.lua
```diff
@@ -109,4 +109,21 @@ QBConfig.Server.Discord = "" -- Discord invite link
...
+    ['phone_invoices'] = 'citizenid',
+    ['phone_messages'] = 'citizenid',
+    ['playerskins'] = 'citizenid',
+    ['player_contacts'] = 'citizenid',
+    ['player_houses'] = 'citizenid',
+    ['player_mails'] = 'citizenid',
+    ['player_outfits'] = 'citizenid',
+    ['player_vehicles'] = 'citizenid',
+    -- ['bank_accounts_new'] = 'creator'
+}
```
Manason, line 127: Could we put this in a new Config object? I'm not sure there is a reason to expose this table data outside of the core.

Also consider putting a comment that these are the rows that will be deleted when a character is deleted.

Also consider renaming to CharacterDataTables. Using character as much as possible will open us up to one day separating the concept of a player

Manason (changes requested):

Manason:
> Yes it solves but I don't think it will be user friendly. As I understand you don't want to write non qbx tables into qbx-core, you can do it by leaving only qbx tables in the cfg I sent. Users who want to add other tables can configure the config file.

Ok. I think I'm convinced since we do want everything in one transaction. See my requested changes, but I'm willing to approve the concept of this PR.

config.lua
```diff
@@ -110,3 +110,21 @@ QBConfig.Server.CheckDuplicateLicense = true -- Check for duplicate rockstar lic
 QBConfig.Server.Permissions = { 'god', 'admin', 'mod' } -- Add as many groups as you want here after creating them in your server.cfg
 
 QBConfig.NotifyPosition = 'top-right' -- 'top' | 'top-right' | 'top-left' | 'bottom' | 'bottom-right' | 'bottom-left'
+
+---@alias TableName string
+---@alias ColumnName string
+---@type table<TableName, ColumnName>
+CharacterDataTables = {
```
Manason, line 117: Config = {}
Config.CharacterDataTables = ...

I like having the Config object to signify that this isn't supposed to be a mutable global variable, but rather an immutable field set in config

Manason (changes requested):

Manason (approved):

PR #133, opened 2023-09-15
REPRO NEEDED: Random death upon login?

mockdot:
## Description

This is an attempt to fix the random scenario where you login and your character is dead. I have not been able to properly pin-point where it is being caused, after looking at some code and I guess plain guessing, I believe it might be related to the way that hunger and thirst are working. Is it possible that when playerDropped is executed we are not saving it properly and even if player is logged out it continues to go down on the server? If we look at logout, we are using it: https://github.com/Qbox-project/qbx-core/blob/8b159d595d275aa38e39efe6cd20fe8cf3f6d19c/server/player.lua#L163

Would love to see if more people can test and reproduce.

## Checklist

### DIFF
--- a/client/character.lua
+++ b/client/character.lua
@@ -212,6 +212,7 @@ local function createCharacter(cid)
         cid = cid
     })
     destroyPreviewCam()
+    TriggerServerEvent('qb-cityhall:server:getIDs')
     return true
 end
 
--- a/client/character.lua
+++ b/client/character.lua
@@ -212,7 +212,6 @@ local function createCharacter(cid)
         cid = cid
     })
     destroyPreviewCam()
-    TriggerServerEvent('qb-cityhall:server:getIDs')
     return true
 end
 
--- a/server/events.lua
+++ b/server/events.lua
@@ -31,7 +31,19 @@ AddEventHandler('playerDropped', function(reason)
     if not QBCore.Players[src] then return end
     GlobalState.PlayerCount -= 1
     local Player = QBCore.Players[src]
+    if not Player then return end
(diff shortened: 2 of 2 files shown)


Manason:
If that's the case, wouldn't it mean the root cause of this bug is players who dropped aren't being properly removed from the player list that is using by the loop to drop hunger/thirst? Shouldn't we look there for the fix?

mockdot:
You are completely right, I did not pay attention that the cron actually does have a player table, it's possible this isn't being properly deleted upon player dropped I guess? https://github.com/Qbox-project/qbx-core/blob/8b159d595d275aa38e39efe6cd20fe8cf3f6d19c/server/loops.lua#L1

Manason:
> You are completely right, I did not pay attention that the cron actually does have a player table, it's possible this isn't being properly deleted upon player dropped I guess?
> 
> https://github.com/Qbox-project/qbx-core/blob/8b159d595d275aa38e39efe6cd20fe8cf3f6d19c/server/loops.lua#L1

Check where player is removed from QBCore.Players on drop event.

mockdot:
Seems to be along here: https://github.com/Qbox-project/qbx-core/blob/8b159d595d275aa38e39efe6cd20fe8cf3f6d19c/server/events.lua#L37

mockdot:
Doing some debugprint tests, it does appear to put it to nil when closing the game and the QBCore.Players table does not show that user anymore on cron execution... not sure if it's a race condition

BerkieBb:
That indentation is a little off, unless it's just my phone

mockdot:
I think this might have been fixed by the changes done with qbx-medical/qbx-hud.

PR #140, opened 2023-09-17
Player save position (properties)

TonybynMp4:
## Description

Prevents saving the position when inside of a property/apartment, (make sure to save player before entering ofc)

## Checklist

### DIFF
--- a/server/player.lua
+++ b/server/player.lua
@@ -557,9 +557,12 @@ end
 ---Save player info to database (make sure citizenid is the primary key in your database)
 ---@param source Source
 function QBCore.Player.Save(source)
-    local ped = GetPlayerPed(source)
-    local pcoords = GetEntityCoords(ped)
+     local ped = GetPlayerPed(source)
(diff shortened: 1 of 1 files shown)


Manason:
If someone doesn't have that player state, looks like this will have the same behavior as before correct saving player position wherever they may be, so it seems backwards compatible.

I'm guessing the player position shouldn't be saved within a property because it won't load and the player will fall through world?

TonybynMp4:
> I'm guessing the player position shouldn't be saved within a property because it won't load and the player will fall through world?

There's two ways you can go about this,
either you have the janky qb-houses way that saves inside the property, wich means you need to always have the shell in the same coords (for IPLs it wouldn't be that problematic i guess), you also need to store wich property etc in db, and yes it can be buggy.

or you just spawn the player outside the property and avoid any problems related to that (gameplay-wise that doesn't change much anyway tbh)
This avoids having to store ANYTHING related to properties in the database and is just way simpler to implement and maintain :)

Manason (approved):

PR #141, opened 2023-09-17
refactor: ids and licenses

mockdot:
## Description

- IDs are now given from qbx-cityhall using um-idcard
- giveStarterItems function is now only used to give starter items defined in QBCore.Shared.StarterItems

## Checklist


client/character.lua
```diff
@@ -212,6 +212,7 @@ local function createCharacter(cid)
         cid = cid
     })
     destroyPreviewCam()
+    TriggerServerEvent('qb-cityhall:server:getIDs')
```
Manason: Is this needed? I thought qbx-cityhall was consuming the character creation event to give ids?

shared/main.lua
```diff
@@ -2,8 +2,6 @@ QBShared = QBShared or {}
 
 QBShared.StarterItems = {
     phone = { amount = 1, item = 'phone' },
-    id_card = { amount = 1, item = 'id_card' },
-    driver_license = { amount = 1, item = 'driver_license' },
 }
```
Manason, line 5: Should we move this to config? It doesn't really seem appropriate as part of QBShared

Manason (changes requested):

Manason (approved):

shared/main.lua
mockdot: I think that's a good idea, let me go ahead and move it

mockdot:
I have not been able to test it yet, getting black screen when creating chars, but trying to show the idea

config.lua
```diff
@@ -132,3 +129,9 @@ Config.CharacterDataTables = {
     ['player_outfits'] = 'citizenid',
     ['player_vehicles'] = 'citizenid',
 } -- Rows to be deleted when the character is deleted
+
+Config.StarterItems = { -- Character starting items
+    phone = { amount = 1, item = 'phone' },
+    id_card = { amount = 1, item = 'id_card', type = 'id'},
+    driver_license = { amount = 1, item = 'driver_license', type = 'license'},
+}
```
Manason, line 137: This should be an array

config.lua
```diff
@@ -132,3 +129,9 @@ Config.CharacterDataTables = {
     ['player_outfits'] = 'citizenid',
     ['player_vehicles'] = 'citizenid',
 } -- Rows to be deleted when the character is deleted
+
+Config.StarterItems = { -- Character starting items
+    phone = { amount = 1, item = 'phone' },
+    id_card = { amount = 1, item = 'id_card', type = 'id'},
+    driver_license = { amount = 1, item = 'driver_license', type = 'license'},
```
Manason, line 136: Needs a field for metadata? What is type?

server/character.lua
```diff
@@ -6,10 +6,53 @@ local function getAllowedAmountOfCharacters(license2, license)
...
+    for _, v in pairs(Config.StarterItems) do
+        if v.type == 'id' then
+            local metadata = {
+                cardtype = v,
+                citizenid = Player.PlayerData.citizenid,
+                firstname = Player.PlayerData.charinfo.firstname,
+                lastname = Player.PlayerData.charinfo.lastname,
+                birthdate = Player.PlayerData.charinfo.birthdate,
+                sex =  getSexString(Player.PlayerData.charinfo.gender),
+                nationality = Player.PlayerData.charinfo.nationality,
```
Manason, line 33: metadata should be in the config itself, even if that means putting a function in the config

Manason (changes requested):

Manason:
Where is money added? We should probably move that to this as well since it's an item in ox_inventory

config.lua
```diff
@@ -129,3 +129,55 @@ Config.CharacterDataTables = {
...
+        nationality = PlayerData.charinfo.nationality,
+        mugShot = 'none',
+    }
+    return metadata
+end
+
+---@alias ItemName string
+---@alias ItemAmount number
+---@alias ItemMetadata table
+---@type table<ItemName, ItemAmount, ItemMetadata>
```
Manason, line 178:
```suggestion
---@type { item: string, amount: integer, metadata: table|function }
```
I don't believe you can specify a table<> with more than two values

config.lua
```diff
@@ -129,3 +129,55 @@ Config.CharacterDataTables = {
...
+Config.StarterItems = { -- Character starting items
+    { item = 'phone', amount = 1 },
+    { item = 'id_card', amount = 1, metadata = getItemMetadata('id_card', 'id')},
```
Manason, line 181: This needs to be an in-line function, as it cannot be evaluated when the file loads, but instead needs to be evaluated when the item is being added to the character.

Manason (changes requested):

config.lua
mockdot: You are referring to using getItemMetadata()? I should specify the metadata directly instead of using the function?

mockdot:
Seems to work great!

config.lua
```diff
@@ -129,3 +129,10 @@ Config.CharacterDataTables = {
     ['player_outfits'] = 'citizenid',
     ['player_vehicles'] = 'citizenid',
 } -- Rows to be deleted when the character is deleted
+
+---@type { name: string, amount: integer, metadata: table | fun(source: number): table}
+Config.StarterItems = { -- Character starting items
+    { name = 'phone', amount = 1 },
+    { name = 'id_card', amount = 1, metadata = function(source) exports['um-idcard']:GetMetaLicense(source, {'id_card'}) end },
+    { name = 'driver_license', amount = 1, metadata = function(source) exports['um-idcard']:GetMetaLicense(source, {'driver_license'}) end },
```
BerkieBb, line 137: I think you forgot the `return` in both functions

BerkieBb (changes requested):

Manason:
Awaiting new release of um-idcard to approve this. We should also add a lib.checkVersion within the metadata function to provide extra verification that people are running um-idcard

config.lua
```diff
@@ -129,3 +129,10 @@ Config.CharacterDataTables = {
     ['player_outfits'] = 'citizenid',
     ['player_vehicles'] = 'citizenid',
 } -- Rows to be deleted when the character is deleted
+
+---@type { name: string, amount: integer, metadata: table | fun(source: number): table}
+Config.StarterItems = { -- Character starting items
+    { name = 'phone', amount = 1 },
+    { name = 'id_card', amount = 1, metadata = function(source) if not lib.checkDependency('um-idcard', '1.2.8') then return else exports['um-idcard']:GetMetaLicense(source, {'id_card'}) end end },
```
Manason, line 136:
```suggestion
    { name = 'id_card', amount = 1, metadata = function(source) if not lib.checkDependency('um-idcard', '1.2.8', true) then error() end exports['um-idcard']:GetMetaLicense(source, {'id_card'}) end end },
```
Prints and errors the dependency. We want to fail loud here.

Manason (changes requested):

config.lua
```diff
@@ -129,3 +129,10 @@ Config.CharacterDataTables = {
     ['player_outfits'] = 'citizenid',
     ['player_vehicles'] = 'citizenid',
 } -- Rows to be deleted when the character is deleted
+
+---@type { name: string, amount: integer, metadata: table | fun(source: number): table}
+Config.StarterItems = { -- Character starting items
+    { name = 'phone', amount = 1 },
+    { name = 'id_card', amount = 1, metadata = function(source) if not lib.checkDependency('um-idcard', '1.2.8', true) then error() end exports['um-idcard']:GetMetaLicense(source, {'id_card'}) end end },
+    { name = 'driver_license', amount = 1, metadata = function(source) if not lib.checkDependency('um-idcard', '1.2.8') then return else exports['um-idcard']:GetMetaLicense(source, {'driver_license'}) end end },
```
Manason, line 137:
```suggestion
    { name = 'driver_license', amount = 1, metadata = function(source) if not lib.checkDependency('um-idcard', '1.2.8', true) then error() end exports['um-idcard']:GetMetaLicense(source, {'driver_license'}) end end },
```

Manason (changes requested):

PR #167, opened 2023-09-25
chore: remove qbx-core GetCoreObject export in favor of module

Manason:
## Description

- Removes exports['qbx-core']:GetCoreObject() as people should just be using the module
- exports['qb-core']:GetCoreObject() will continue to work for those needing backwards compatibility and can be used at the same time as a module to switch a resource over gradually off of the qb core object
- Also renaming the global variable supplied by the module to 'QbxCore' so that resources that currently contain QBCore variables won't have a conflict, and both can be used at the same time. The rename also helps communicate that QbxCore does not contain the same functions that QBCore does.

## Checklist

### DIFF
--- a/client/main.lua
+++ b/client/main.lua
@@ -5,8 +5,3 @@ QBCore.Shared = require 'shared.main'
 QBCore.Functions = require 'client.functions'
 
 IsLoggedIn = false
-
----@deprecated import QBCore using module 'qbx-core:core' https://qbox-docs.vercel.app/resources/core/import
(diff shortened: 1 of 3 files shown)


BerkieBb:
QbxCore doesn't sound so great tbh, why not something better whilst we're at it?

Manason:
> QbxCore doesn't sound so great tbh, why not something better whilst we're at it?

Open to suggestions

BerkieBb:
QBXCore

PR #174, opened 2023-09-26
refactor: deprecate bucket wrappers

Manason:
## Description

- moved bucket fields of core object and function to bridge module
- made bridge core a global so that it can be referenced in core code to maintain compatibility from events and such


Manason:
After discussion, we decided not to do this. Will put up a follow-up CL to deprecate exposing them via core object.

PR #184, opened 2023-10-03
feat: default spawn to last location

### LINKED ISSUE #164: spawn last location when qbx-spawn isnt present
### The problem

When qbx-spawn isn't present, it spawns you at a set coord even if its with a character that's already made

### Ideal solution

Make it so when qbx-spawn is removed and you select a made character, it spawns it at the last location.

### Alternative solutions

no

### Additional context

no

Manason:
- fallback to spawn last location if no other spawn resources are found fixes #164 
- fix rename of qbx-spawn to qbx_spawn


xrayn22:
this should be a config option to whether it should change to the configured default spawn or last spawn location @Manason

PR #194, opened 2023-10-08
player buckets fix

DevX32:
## Checklist


FjamZoo (approved):

server/events.lua
```diff
@@ -1,5 +1,5 @@
 -- Event Handler
-
+QBX.Player_Buckets = {}
```
Manason, line 2: These are already defined here: https://github.com/Qbox-project/qbx_core/blob/main/server/main.lua#L15

Why do they need to be set here as well?

Manason (changes requested):

server/events.lua
DevX32: when player drops error pop ups player_buckets nil

Manason: I'm afraid I don't understand.

DevX32: if u are entering the server and u exited or something u get and error called playerbuckets nil

Manason: This doesn't seem like the right solution as one of the two QBX.Player_Buckets = {} may overwrite existing data and it's confusing to clear it twice. Rather instead I would suggest checking that QBX.Player_Buckets exists before trying to clear the player out of one

Manason (changes requested):

server/events.lua
DevX32: i will  try to find the perma and right solution

PR #195, opened 2023-10-09
minor changes on character selection screen

hakanesnn:
getting active skin & random ped instead of random clothes & gang information to character infos & reorder character information

PR #196, opened 2023-10-09
fixing the giving of starter items

hakanesnn:
items are not given because inventory is not created without playerloaded

### DIFF
--- a/server/character.lua
+++ b/server/character.lua
@@ -16,7 +16,7 @@ local function giveStarterItems(source)
     end
 
     if not invCreated then return error('starting items could not be given because no inventory could be found') end
-    
+
     for i = 1, #Config.StarterItems do
         local item = Config.StarterItems[i]
         if item.metadata and type(item.metadata) == 'function' then
--- a/client/character.lua
+++ b/client/character.lua
@@ -3,6 +3,19 @@ if Config.Characters.UseExternalCharacters then return end
 local previewCam = nil
 local randomLocation = Config.Characters.Locations[math.random(1, #Config.Characters.Locations)]
 
+local randomPedModels = {
+    `a_m_o_soucent_02`,
+    `mp_g_m_pros_01`,
+    `a_m_m_prolhost_01`,
+    `a_f_m_prolhost_01`,
+    `a_f_y_smartcaspat_01`,
+    `a_f_y_runner_01`,
+    `a_f_y_vinewood_04`,
+    `a_f_o_soucent_02`,
+    `a_m_y_cyclist_01`,
+    `a_m_m_hillbilly_02`,
+}
+
 local function setupPreviewCam()
     DoScreenFadeIn(1000)
     SetTimecycleModifier('hud_def_blur')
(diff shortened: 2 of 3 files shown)


Manason (commented):
Has this been tested? I would think this would need to be in a thread to avoid blocking further execution.

Manason:
> > Has this been tested? I would think this would need to be in a thread to avoid blocking further execution.
> 
> Yes, I tested it.

And you got an id card? Why doesn't this need to be in a thread?

BerkieBb:
> > > Has this been tested? I would think this would need to be in a thread to avoid blocking further execution.
> > 
> > 
> > Yes, I tested it.
> 
> And you got an id card? Why doesn't this need to be in a thread?

It already is, it's always called in a callback, which is basically an event which always runs in a thread

Manason:
> > > > Has this been tested? I would think this would need to be in a thread to avoid blocking further execution.
> > > 
> > > 
> > > Yes, I tested it.
> > 
> > 
> > And you got an id card? Why doesn't this need to be in a thread?
> 
> It already is, it's always called in a callback, which is basically an event which always runs in a thread

It would block the callback return, which triggers the event that causes the inventory to load? Something isn't right here.

BerkieBb:
> > > > > Has this been tested? I would think this would need to be in a thread to avoid blocking further execution.
> > > > 
> > > > 
> > > > Yes, I tested it.
> > > 
> > > 
> > > And you got an id card? Why doesn't this need to be in a thread?
> > 
> > 
> > It already is, it's always called in a callback, which is basically an event which always runs in a thread
> 
> It would block the callback return, which triggers the event that causes the inventory to load? Something isn't right here.

It would not block anything, it's just holding up the thread like it's supposed to

BerkieBb:
Also, the login function is called before giving starter items so the inventory will load fine

server/character.lua
```diff
@@ -6,6 +6,10 @@ end
 
 ---@param source Source
 local function giveStarterItems(source)
+    while not exports.ox_inventory:GetInventory(source) do
+        Wait(100)
+    end
+
```
Manason, line 12: Use https://overextended.dev/ox_lib/Modules/WaitFor/Shared

Manason (changes requested):

server/character.lua
BerkieBb: We want waitfor in there because it makes sure an infinite loop is not possible using the timeout, try to get it to work and if you really can't, we'll get an alternative

BerkieBb: You can mimick the code of waitFor to wait longer and to check using `not` instead of nil explicitly in here

Manason: An alternative solution would be to register a net event for https://overextended.dev/ox_inventory/Events/Client#ox_inventoryupdateinventory on the client and use it to trigger a custom qbox server event which stores a global value that the inventory is loaded, and then you can wait for that global variable.

server/character.lua
```diff
@@ -6,6 +6,17 @@ end
...
 ---@param source Source
 local function giveStarterItems(source)
+    local getInv = function(source) return exports.ox_inventory:GetInventory(source) ~= false end
+    local i = 0
+    local invCreated = getInv(source)
+    while not invCreated and i < 100 do
+        i += 1
+        Wait(100)
+        invCreated = getInv(source)
+    end
```
Manason, line 16: repeat until? Or do the event based method

Manason (changes requested):

ChatDisabled:
Also check if everything is as intended after the merger of https://github.com/Qbox-project/qbx_core/pull/195 and if I properly resolved conflicts without accidently removing anything.

server/character.lua
Manason: When I say event based method I'm referring to https://github.com/Qbox-project/qbx_core/pull/196#discussion_r1351502762

PR #198, opened 2023-10-10
add Progressbar

StrixLuca:
I know that you can get the progress bar from ox_lib, but this is just better, not everyone looks at the ox docs and a core does need a progress bar ps not hate me xd

### DIFF
--- a/client/functions.lua
+++ b/client/functions.lua
@@ -35,6 +35,42 @@ end
 
 exports('Notify', Notify)
 
+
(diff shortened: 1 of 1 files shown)


StrixLuca:
for the noob devs is good ahah

StrixLuca:
any news

Manason:
I'm not sure that we want a wrapper on progress bar. We support such a wrapper via the bridge already. I don't understand why a core needs a progress bar, nor what not everyone looks at the docs has to do with it. They'd either have to look at Qbox or Ox docs, or just read the code. I'm failing to understand why this is better than using the ox_lib progress bar.

PR #199, opened 2023-10-11
Spawn Selector & qbx_core:client:playerLoaded & qbx_core:server:playerLoaded

hakanesnn:
## Description
- Spawn selector does the same thing as qbx_spawn but it is interfaced with ox_lib context menu, I used `LoadResourceFile` to get the qbx_apartments config, maybe it can be changed to be requested with export.
- Can be started in owned houses or apartments, this feature can be disabled in the config file.
- Added `qbx_core:server:playerLoaded` and `qbx_core:client:playerLoaded` event, triggers from server after character selection, returns playerData as parameter to client event, triggers after player creation, triggers old qbcore events from client for compatibility.
- When a new character is created, the inventory will be loaded because the `giveStarterItems` function will be called after the playerloaded is triggered from the server, I did not delete the loop because I did not test this in detailed, you can try and delete the loop.
- I changed the string `Qbox Multichar` to `Character List` in the English language file, it seems more logical for players.

Tested by installing the latest txadmin recipe and deleting qbx_spawn.

## Checklist

### DIFF
--- a/client/spawn.lua
+++ b/client/spawn.lua
@@ -100,8 +100,6 @@ lib.callback.register('qbx_core:client:getSpawnLocation', function(data)
         promise:resolve(false)
     end
 
-    :: skip ::
-
     return Citizen.Await(promise)
 end)
 


client/events.lua
```diff
@@ -1,10 +1,18 @@
...
     ShutdownLoadingScreenNui()
+    QBX.PlayerData = val
     QBX.IsLoggedIn = true
     if not Config.Server.PVP then return end
     SetCanAttackFriendly(cache.ped, true, false)
     NetworkSetFriendlyFireOption(true)
+
+    --backward compatibility
+    TriggerEvent('QBCore:Client:OnPlayerLoaded')
+    TriggerServerEvent('QBCore:Server:OnPlayerLoaded')
```
Manason, line 15: What's the benefit of naming a new event if we'll need to continue to trigger these for backwards compatibility reasons? Why not continue to use the same event name?

client/spawn.lua
```diff
@@ -0,0 +1,164 @@
+local cam
+lib.callback.register('qbx_core:client:getSpawnLocation', function(data)
+    SetEntityVisible(cache.ped, false)
+    SetCloudHatOpacity(0.0)
+    cam = CreateCamWithParams('DEFAULT_SCRIPTED_CAMERA', Config.DefaultSpawn.x, Config.DefaultSpawn.y, Config.DefaultSpawn.z + 1500, -85.00, 0.0, 0.0, 100.00, false, 0)
+    SetCamActive(cam, true)
+    RenderScriptCams(true, false, 1, true, true)
+    DoScreenFadeIn(500)
```
Manason, line 8: I wouldn't expect getSpawnLocation to mutate any states. Consider renaming.

Manason (changes requested):
Started reviewing, but realize I actually want to talk about high level design first before getting into the weeds here. Core should have 0 dependencies on other Qbox resources, and a bias against Config enabled features. My question is could we design this in a way that common spawn functionality such as destroying a preview cam, making an entity visible etc, is done in core, but we trigger an event that can be consumed by houses, apartments, etc to take additional steps specific to spawning there? I'm concerned that too much of other resources is now leaking into the core with this PR.

PR #200, opened 2023-10-12
feat: GetLocations export

hakanesnn:


Manason:
What do you need these locations for outside of core?

Manason:
> > What do you need these locations for outside of core?
> 
> I want to use them to teleport in the admin menu, they are written in core but they are not used anywhere.

Oh. Should they be moved into qbx_adminmenu then if they aren't being used in core?

Manason (approved):

Issue #183, opened 2023-09-28
Add export GetVehiclesByCategory()


Manason:
### The problem

It's a common operation to create a map of categories to vehicles. adminmenu & vehicleshop both do it. This results in duplicate code across resources implementing the same feature

### Ideal solution

create a centralized core export to reduce duplicate code in other resources for a common operation.

### Alternative solutions

add a utility function to the utility module which converts the vehicles to the desired format. If we go this route maybe we could make a generic remapper function which inputs the key to map on so it can be re-used to create a map of hashes to vehicles, or for other tables as well.

At the very least, if going the module route, deprecate GetVehiclesByName and GetVehiclesByHash and create GetVehicles

TonybynMp4:
So i guess, if i understand what you'd like correctly, that you'd want a function like this?
`
---@param vehicle model | hash
---@param category string
GetVehicles(vehicle, category)
`

Manason:
> So i guess, if i understand what you'd like correctly, that you'd want a function like this?
> 
> ```lua
> ---@param vehicle model | hash
> ---@param category string
> function GetVehicles(vehicle, category)
> ...
> end)
> ```

Negative. I mean a function that returns all the vehicles, mapped by category so that I can do exports.qbx_core:GetVehiclesByCategory()[category]

TonybynMp4:
I'm talking about that 🤔 
> At the very least, if going the module route, deprecate GetVehiclesByName and GetVehiclesByHash and create GetVehicles

Manason:
> I'm talking about that 🤔
> 
> > At the very least, if going the module route, deprecate GetVehiclesByName and GetVehiclesByHash and create GetVehicles

Given that we already store this mapping in core, I think the export route is preferred over the module solution.

PR #206, opened 2023-10-19
Add GetVehiclesByCategory() Export

TonybynMp4:
## Description
Adds an export that returns a mapped table of shared/vehicles.lua by category.
Answers #183 

### exports.qbx_core:GetVehiclesByCategory().cycles
will return all the biccyles, one example in this table:
```lua
 {
    "brand": "Bike",
    "category": "cycles",
    "hash": -400295096,
    "model": "tribike3",
    "name": "Tri Bike 3",
    "price": 520,
    "shop": "pdm"
}
```

## Checklist

### DIFF
--- a/client/main.lua
+++ b/client/main.lua
@@ -31,6 +31,8 @@ end
 
 exports('GetVehiclesByHash', GetVehiclesByHash)
 
+exports('GetVehiclesByCategory', GetVehiclesByCategory)
+
 ---@return table<number, Weapon>
 function GetWeapons()
     return QBX.Shared.Weapons
--- a/client/main.lua
+++ b/client/main.lua
@@ -31,6 +31,12 @@ end
 
 exports('GetVehiclesByHash', GetVehiclesByHash)
 
+---@param vehicles table<string, Vehicle>
+---@return table<string, table<number, Vehicle>>
+function GetVehiclesByCategory(vehicles)
+	return MapTableBySubfield('category', vehicles)
+end
+
 exports('GetVehiclesByCategory', GetVehiclesByCategory)
 
 ---@return table<number, Weapon>
--- a/client/main.lua
+++ b/client/main.lua
@@ -31,10 +31,9 @@ end
 
 exports('GetVehiclesByHash', GetVehiclesByHash)
 
----@param vehicles table<string, Vehicle>
 ---@return table<string, Vehicle[]>
-function GetVehiclesByCategory(vehicles)
-	return MapTableBySubfield('category', vehicles)
+function GetVehiclesByCategory()
+	return MapTableBySubfield('category', QBX.Shared.Vehicles)
 end
 
 exports('GetVehiclesByCategory', GetVehiclesByCategory)
--- a/client/main.lua
+++ b/client/main.lua
@@ -32,7 +32,7 @@ end
 exports('GetVehiclesByHash', GetVehiclesByHash)
 
 ---@param vehicles table<string, Vehicle>
----@return table<string, table<number, Vehicle>>
+---@return table<string, Vehicle[]>
 function GetVehiclesByCategory(vehicles)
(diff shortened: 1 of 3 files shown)


modules/utils.lua
```diff
@@ -142,6 +142,21 @@ function GenerateRandomPlate(pattern) -- luacheck: ignore
...
+	for vehicleIndex, vehicleData in pairs(Vehicles) do
+		if not mappedVehicles[vehicleData.category] then
+			mappedVehicles[vehicleData.category] = {}
+		end
+
+		mappedVehicles[vehicleData.category][vehicleIndex] = vehicleData
+	end
+	return mappedVehicles
+end
+
```
Manason, line 159: This should not be a module function. Move this to where GetVehiclesByHash is calculated as this is similar; a transform of the original data table. Consistency is important, if we're going to do transforms as module functions we should do them for vehicles by hash as well. A good general util function might be to create a mapping of any common subfield for any table..

Manason (changes requested):

modules/utils.lua
TonybynMp4: better now? :)

modules/utils.lua
```diff
@@ -142,6 +142,26 @@ function GenerateRandomPlate(pattern) -- luacheck: ignore
...
 end
 
+--- maps a table by the given subfield
+--- @param subfield string
+--- @param table table
+--- @return table
+function MapTableBySubfield(subfield, table)
+	local mappedTable = {}
+	for tableIndex, tableData in pairs(table) do
+		if not tableData[subfield] then
```
Manason, line 152: store tableData[subfield] in a variable. It's used several times and will increase readability to extract it.

client/main.lua
```diff
@@ -31,6 +31,14 @@ end
 
 exports('GetVehiclesByHash', GetVehiclesByHash)
 
+---@param vehicles table<string, Vehicle>
+---@return table<string, table<number, Vehicle>>
+function GetVehiclesByCategory(vehicles)
```
Manason, line 36: Does this need a parameter?

client/main.lua
```diff
@@ -31,6 +31,14 @@ end
 
 exports('GetVehiclesByHash', GetVehiclesByHash)
 
+---@param vehicles table<string, Vehicle>
+---@return table<string, table<number, Vehicle>>
```
Manason, line 35:
```suggestion
---@return table<string, Vehicle[]>
```

Manason (changes requested):

client/main.lua
TonybynMp4: As discussed on discord, no it doesn't.

modules/utils.lua
```diff
@@ -142,6 +142,26 @@ function GenerateRandomPlate(pattern) -- luacheck: ignore
...
+			goto continue
+		end
+		if not mappedTable[tableData[subfield]] then
+			mappedTable[tableData[subfield]] = {}
+		end
+
+		mappedTable[tableData[subfield]][tableIndex] = tableData
```
Manason, line 159: What happens if table is a map rather than an array?

TonybynMp4: I guess it would return itself?

modules/utils.lua
```diff
@@ -145,18 +145,19 @@ end
 --- maps a table by the given subfield
 --- @param subfield string
 --- @param table table
---- @return table
+--- @return table<string, table>
```
Manason, line 148:
```suggestion
--- @return table<any, table[]>
```

modules/utils.lua
TonybynMp4: If what you mean is can i map a table, and then re-map that, the answer is no.
since we're mapping arrays, there wouldn't be any subfield to select.

Manason (approved):

PR #223, opened 2023-11-03
fix(server/character): spawning players in different buckets

Manason:
- Investigated reports of players spawning in and not being able to see each other (being in different buckets)
- Observed that SetPlayerRoutingBucket doesn't seem to be consistently working. Sometimes it sets the routing bucket and other times it doesn't. It seemed to only work when multiple players connected at the same time, but this might just be a coincidence. When this actually worked to set the bucket to the player source, the second call to SetPlayerRoutingBucket to 0 had no effect. Unsure why this is, but best guess is that there is something broken about SetPlayerRoutingBucket native.
- Removed code setting the player's bucket to source. This has the effect of player's being able to see each other when in the same character selection location. However, this is preferable to players spawning in different buckets.

### DIFF
--- a/server/character.lua
+++ b/server/character.lua
@@ -39,23 +39,6 @@ lib.callback.register('qbx_core:server:getPreviewPedData', function(_, citizenId
     return ped.skin, ped.model and joaat(ped.model)
 end)
 
-AddEventHandler('playerJoining', function()
-    SetPlayerRoutingBucket(source, source)
-end)
-
-AddEventHandler('onResourceStart', function(resourceName)
-    if resourceName ~= GetCurrentResourceName() then return end
-    Wait(100)
-    local players = GetPlayers()
-    for i = 1, #players do
-        local playerId = players[i]
-        local playerIdNum = tonumber(playerId)
-        if playerIdNum then
-            SetPlayerRoutingBucket(playerId, playerIdNum)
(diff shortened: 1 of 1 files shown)


BerkieBb:
We could also delay setting the routing bucket to right before the UI is shown, maybe that'll help

Manason:
> We could also delay setting the routing bucket to right before the UI is shown, maybe that'll help

I tried that. I experimented with a server callback when the client's network session starts, which is the trigger to show the UI. Also when spawning, the bucket wasn't being set to 0, which is well after any loading. One thing I did not try is waiting until the model is first set to do it.

In any case, I think the best move forward is to merge this PR to mitigate the more severe bug, and then undertake a separate effort to deal with trying to re-introduce buckets.

BerkieBb:
> > We could also delay setting the routing bucket to right before the UI is shown, maybe that'll help
> 
> I tried that. I experimented with a server callback when the client's network session starts, which is the trigger to show the UI. Also when spawning, the bucket wasn't being set to 0, which is well after any loading. One thing I did not try is waiting until the model is first set to do it.
> 
> In any case, I think the best move forward is to merge this PR to mitigate the more severe bug, and then undertake a separate effort to deal with trying to re-introduce buckets.

Alright

Issue #219, opened 2023-11-02
qbx_core failed to load in the server


gringoDiaz:
### Summary

qbx_core error

### Reproduction

Every time you login to the server qbx_core default character selection not showing up and spawn you as a ped then have error says failed to load model

### Expected behavior

Error 

@qbx_core/client/character.lua:387: failed to load model '279228114' (waited 8.6ms)

### Actual behavior

You load the server as a ped not in your default character

### Additional context

![image](https://github.com/Qbox-project/qbx_core/assets/63420825/a5f98451-cbf0-4d36-afa8-28f906fd57a4)

### Last Updated

today

### Custom Resources

no

### Resource Rename

no

SKITTLE6969:
@gringoDiaz could you try my PR and let me know asap?

gringoDiaz:
@SKITTLE6969 I've try the PR and it fixed the error but i takes time for the model to load

SKITTLE6969:
> @SKITTLE6969 I've try the PR and it fixed the error but i takes time for the model to load

you can reduce the time from the Config.LoadingModelsTimeout = 10000 to something lower

gringoDiaz:
Yes, it helps i tried to lower it to 5000

PR #224, opened 2023-11-03
adding LoadingModelsTimeout

SKITTLE6969:
Waiting time for ox_lib to load the models before throws an error, for low specs pc

should fix https://github.com/Qbox-project/qbx_core/issues/219

## Checklist

- [✔] I have personally loaded this code into an updated Qbox project and checked all of its functionality.
- [✔] My code fits the style guidelines.
- [✔] My PR fits the contribution guidelines.

### DIFF
--- a/bridge/qb/client/functions.lua
+++ b/bridge/qb/client/functions.lua
@@ -136,7 +136,7 @@ function functions.SpawnVehicle(model, cb, coords, isnetworked, teleportInto)
     if not IsModelInCdimage(model) then return end
 
     isnetworked = isnetworked == nil or isnetworked
-    lib.requestModel(model, Config.LoadingModelsTimeout)
+    lib.requestModel(model)
     local veh = CreateVehicle(model, coords.x, coords.y, coords.z, coords.w, isnetworked, false)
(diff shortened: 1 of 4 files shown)


Manason:
If this is indeed the issue, I think we should try to make a PR to ox_lib to increase the default timeout instead. Loading models using ox_lib is done everywhere in many resources. I'm a bit suspicious of this being the culprit though, as like I said, ox_lib loading models is used in many different resources. I also would like Gringo to test with the increased timeout and confirm that is indeed the issue.

SKITTLE6969:
> If this is indeed the issue, I think we should try to make a PR to ox_lib to increase the default timeout instead. Loading models using ox_lib is done everywhere in many resources. I'm a bit suspicious of this being the culprit though, as like I said, ox_lib loading models is used in many different resources. I also would like Gringo to test with the increased timeout and confirm that is indeed the issue.

this was happening after a recent ox_lib update (before 3 or 4 recent updated ago) and I have av_multicharacter which uses ox_lib to request models. so any lib release before Aug something was working without any issue and only the latest ones are having this wait issue

ChatDisabled:
Overextended is known to never merge PRs so we will do it. :trollface:

Issue #226, opened 2023-11-04
Players can see each other during character selection


Manason:
### Summary

Players are in the same bucket, so can see and hear each other when in the same character selection location

### Reproduction

1. Remove all character selection locations except for 1
2. Have someone else load into character selection
3. Load in yourself
4. Observe that you can see each other and hear each other

### Expected behavior

You wouldn't be able to see nor hear other players selecting characters

### Actual behavior

You can see and hear other players selecting characters

### Additional context

This is due to not putting players in different buckets on load-in due to a more severe bug associated with that. See https://github.com/Qbox-project/qbx_core/pull/223

### Last Updated

N/A

### Custom Resources

N/A

### Resource Rename

N/A

alberttheprince:
Is this still a bug? This is an unfortunate side effect of bucketing not working correctly correct? And why where the characters are kept and the camera displays the preview character are separate in multichar.

Manason:
> Is this still a bug? This is an unfortunate side effect of bucketing not working correctly correct? And why where the characters are kept and the camera displays the preview character are separate in multichar.

We intentionally turned off bucketing, because it was causing a more severe bug. This is still an open issue affecting those using the built-in multi char.

Manason:
I don't believe this has yet been addressed. #248 was rolled back.

ChatDisabled:
Issue itself of routing buckets not working with oal has been fixed by Cfx https://github.com/citizenfx/fivem/pull/2286

Manason:
> Issue itself of routing buckets not working with oal has been fixed by Cfx [citizenfx/fivem#2286](https://github.com/citizenfx/fivem/pull/2286)

Yes, but we still need to re-implement them.

mafewtm:
Should be fixed as of #445

PR #248, opened 2023-11-15
fix: seeing/hearing other players in multichar

MafewTM:
## Description

Hopefully fixes seeing/hearing other players in multichar as stated in #226 

## Checklist

### DIFF
--- a/client/character.lua
+++ b/client/character.lua
@@ -44,7 +44,7 @@ local function destroyPreviewCam()
     DestroyCam(previewCam, true)
     RenderScriptCams(false, false, 1, true, true)
     FreezeEntityPosition(cache.ped, false)
-    NetworkConcealPlayer(cache.ped, false)
+    NetworkConcealPlayer(cache.playerId, false)
 end
 
 ---@param citizenId? string
@@ -257,7 +257,7 @@ local function chooseCharacter()
     end
 
     FreezeEntityPosition(cache.ped, true)
-    NetworkConcealPlayer(cache.ped, true)
+    NetworkConcealPlayer(cache.playerId, true)
     Wait(1000)
     SetEntityCoords(cache.ped, randomLocation.pedCoords.x, randomLocation.pedCoords.y, randomLocation.pedCoords.z, false, false, false, false)
     SetEntityHeading(cache.ped, randomLocation.pedCoords.w)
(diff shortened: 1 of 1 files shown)


client/character.lua
```diff
@@ -44,6 +44,7 @@ local function destroyPreviewCam()
     DestroyCam(previewCam, true)
     RenderScriptCams(false, false, 1, true, true)
     FreezeEntityPosition(cache.ped, false)
+    NetworkConcealPlayer(cache.ped, false)
```
FjamZoo, line 47: Use cache.playerId and not cache.ped

client/character.lua
```diff
@@ -256,6 +257,7 @@ local function chooseCharacter()
     end
 
     FreezeEntityPosition(cache.ped, true)
+    NetworkConcealPlayer(cache.ped, true)
```
FjamZoo, line 260: Use cache.playerId and not cache.ped

FjamZoo (changes requested):

FjamZoo (approved):

YvanCywan (approved):

FjamZoo:
BTW this should probably have proper testing due to the fact we use OAL which normally needs every single parameter inside of natives filled out to work properly, and it looks like: https://docs.fivem.net/natives/?_0xBBDF066252829606

Has a 3rd parameter, which could cause issues making this "not work"

BerkieBb:
> BTW this should probably have proper testing due to the fact we use OAL which normally needs every single parameter inside of natives filled out to work properly, and it looks like: https://docs.fivem.net/natives/?_0xBBDF066252829606
> 
> Has a 3rd parameter, which could cause issues making this "not work"

That's actually not OAL that requires it, just the vscode extension. OAL only requires you to not use vector unpacking. You can leave last arguments empty because they always have a default value, usually it's false

FjamZoo:
> > BTW this should probably have proper testing due to the fact we use OAL which normally needs every single parameter inside of natives filled out to work properly, and it looks like: https://docs.fivem.net/natives/?_0xBBDF066252829606
> > Has a 3rd parameter, which could cause issues making this "not work"
> 
> That's actually not OAL that requires it, just the vscode extension. OAL only requires you to not use vector unpacking. You can leave last arguments empty because they always have a default value, usually it's false

Sounds good i will merge then

PR #266, opened 2023-11-24
fix: typo for haskeys export

dusan01:
## Description

On a fresh install, you get "No such export HasKeys in resource qb-vehiclekeys" whenever you sit in the vehicle. I guess this was a typo in the last commit, so this fixes that problem.

## Checklist

### DIFF
--- a/config/client.lua
+++ b/config/client.lua
@@ -68,6 +68,6 @@ return {
 
     --- Only used by QB bridge
     hasKeys = function()
-        return exports['qb-vehiclekeys']:HasKeys()
+        return exports['qbx_vehiclekeys']:HasKeys()
     end,
 }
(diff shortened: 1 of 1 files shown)


config/client.lua
```diff
@@ -68,6 +68,6 @@ return {
 
     --- Only used by QB bridge
     hasKeys = function()
-        return exports['qb-vehiclekeys']:HasKeys()
+        return exports['qbx_vehiclekeys']:HasKeys()
```
solareon, line 71:
```suggestion
        return exports.qbx_vehiclekeys:HasKeys()
```
Since there is no dash you can call it directly without the []

solareon (changes requested):

Manason (approved):

MafewTM (approved):

PR #276, opened 2023-12-06
feat: logging backwards compatibility 

solareon:
## Description

Adds logging backwards compatibility and refactors configuration for logger module. Allows users to decide to enable webhook and oxlib logging at the same time vice limiting to only webhooks or oxlib. Updated the params for logging classes and moved swapped the classes around to better reflect what each class uses.

## Checklist

### DIFF
--- a/bridge/qb/server/events.lua
+++ b/bridge/qb/server/events.lua
@@ -1,3 +1,6 @@
+local config = require 'config.server'
+local logger = require 'modules.logger'
+
 -- Vehicles
 RegisterServerEvent('baseevents:enteringVehicle', function(veh, seat, modelName, netId)
     local src = source
@@ -38,4 +41,22 @@ RegisterServerEvent('baseevents:leftVehicle', function(veh, seat, modelName, net
         event = 'Left'
     }
     TriggerClientEvent('QBCore:Client:VehicleInfo', src, data)
+end)
+
+---Compatability event for logging
+---@param name string source of the log. Usually a playerId or name of a script.
+---@param title string the action or 'event' being logged. Usually a verb describing what the name is doing. Example: SpawnVehicle
+---@param message string the message attached to the log
+---@param color? string what color the message should be
+---@param tagEveryone? boolean Whether a role tag should be applied to this log. Uses config variable for logging role
+---@deprecated use logger module from qbx_core
+RegisterNetEvent('qb-log:server:CreateLog', function(name, title, color, message, tagEveryone)
+    logger.log({
+        source = GetInvokingResource() or "qbx_core",
+        webhook = name,
+        event = title,
+        message = message,
+        color = color,
+        tags = tagEveryone and config.logging.role or nil
+    })
 end)
--- a/bridge/qb/server/events.lua
+++ b/bridge/qb/server/events.lua
@@ -42,21 +42,3 @@ RegisterServerEvent('baseevents:leftVehicle', function(veh, seat, modelName, net
     }
     TriggerClientEvent('QBCore:Client:VehicleInfo', src, data)
 end)
-
----Compatability event for logging
----@param name string source of the log. Usually a playerId or name of a script.
----@param title string the action or 'event' being logged. Usually a verb describing what the name is doing. Example: SpawnVehicle
----@param message string the message attached to the log
----@param color? string what color the message should be
----@param tagEveryone? boolean Whether a role tag should be applied to this log. Uses config variable for logging role
----@deprecated use logger module from qbx_core
-RegisterNetEvent('qb-log:server:CreateLog', function(name, title, color, message, tagEveryone)
-    logger.log({
-        source = GetInvokingResource() or "qbx_core",
-        webhook = name,
-        event = title,
-        message = message,
-        color = color,
-        tags = tagEveryone and config.logging.role or nil
-    })
-end)
--- a/bridge/qb/server/events.lua
+++ b/bridge/qb/server/events.lua
@@ -1,6 +1,3 @@
-local config = require 'config.server'
-local logger = require 'modules.logger'
-
 -- Vehicles
 RegisterServerEvent('baseevents:enteringVehicle', function(veh, seat, modelName, netId)
     local src = source
--- a/config/server.lua
+++ b/config/server.lua
@@ -113,19 +113,17 @@ return {
             end
         },
     },
+
+    -- this configuration is for core events only. putting other webhooks here will have no effect
     logging = {
-        enableOxLogging = false,
-        enableWebhooks = false,
         webhook = {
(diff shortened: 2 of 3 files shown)


config/server.lua
```diff
@@ -113,6 +113,20 @@ return {
...
+        webhook = {
+            ['default'] = '', -- default
+            ['joinleave'] = '', -- default
+            ['ooc'] = '', -- default
+            ['anticheat'] = '', -- default
+            ['playermoney'] = '', -- default
+        },
+        role = {
+            '@everyone' --Role to tag for logging data. Roles use <@%roleid> and users/channels are <@userid/channelid>
+        }
```
Manason, line 128: Maybe add a comment that this is for core only and can't be used as central configuration for other resources? Feel like it could be confusing.

Also, do we really want to tag everyone by default for core logs?

config/server.lua
```diff
@@ -113,6 +113,20 @@ return {
...
     },
+    logging = {
+        enableOxLogging = false,
+        enableWebhooks = false,
+        webhook = {
+            ['default'] = '', -- default
+            ['joinleave'] = '', -- default
+            ['ooc'] = '', -- default
+            ['anticheat'] = '', -- default
+            ['playermoney'] = '', -- default
```
Manason, line 124: should default to nil

bridge/qb/server/events.lua
```diff
@@ -38,4 +41,22 @@ RegisterServerEvent('baseevents:leftVehicle', function(veh, seat, modelName, net
...
+---@deprecated use logger module from qbx_core
+RegisterNetEvent('qb-log:server:CreateLog', function(name, title, color, message, tagEveryone)
+    logger.log({
+        source = GetInvokingResource() or "qbx_core",
+        webhook = name,
+        event = title,
+        message = message,
+        color = color,
+        tags = tagEveryone and config.logging.role or nil
+    })
```
Manason, line 61: I've discussed why I'm against having this event in the first place as core doesn't own this event; small resources does. If you do want to keep it, it needs to be disabled by default.

Manason (changes requested):

config/server.lua
solareon: The `@everyone` tag was a carryover from smallresources. I'll leave it as an empty table for people to use if they want to.

config/server.lua
solareon: done

bridge/qb/server/events.lua
solareon: I'll just remove it. Without having a centralized location for the webhook matching it will just confuse folks more than it'll help. I'll just publish a gist that can be linked for someone to drop in if they want it.

modules/logger.lua
```diff
@@ -138,19 +139,18 @@ end
...
 ---Creates a log using either ox_lib logger, discord webhooks depending on convar
 ---@param log Log
 local function createLog(log)
     if log.webhook then
         ---@diagnostic disable-next-line: param-type-mismatch
         discordLog(log)
-    else
-        lib.logger(log.source, log.event, log.message)
     end
+    lib.logger(log.source, log.event, log.message) -- oxlib fails silently if logging isn't setup correctly.
```
Manason, line 153: This assumes that someone will always want to log to ox_lib instead of some logs going to discord and other logs going to ox_lib. I think that's probably a good assumption, but wanted to call it out. No action required.

modules/logger.lua
```diff
@@ -138,19 +139,18 @@ end
 ---@field source string source of the log. Usually a playerId or name of a resource.
 ---@field event string the action or 'event' being logged. Usually a verb describing what the name is doing. Example: SpawnVehicle
 ---@field message string the message attached to the log
----@field webhook? string Discord logs only. url of the webhook this log should send to
+---@field webhook? string Discord logs only. url of the webhook this log should send to. looks up via config
```
Manason, line 142: what does looks up via config mean? It's a bit unclear

modules/logger.lua
```diff
@@ -138,19 +139,18 @@ end
...
 ---@field color? string Discord logs only. what color the message should be
----@field tags? string[] Discord logs only. tags in discord. Example: ['@admin', '@everyone']
+---@field tags? string[] Discord logs only. tags in discord. Example: {'<@%roleid>', '@everyone'}
 
 ---Creates a log using either ox_lib logger, discord webhooks depending on convar
```
Manason, line 146:
```suggestion
---Logs using ox_lib if ox_lib logging is configured. Additionally logs to discord if a web hook is passed.
```
Wanting to clarify this AND behavior of logging to both.

modules/logger.lua
```diff
@@ -101,12 +101,13 @@ end
 ---@field source string source of the log. Usually a playerId or name of a resource.
 ---@field event string the action or 'event' being logged. Usually a verb describing what the name is doing. Example: SpawnVehicle
 ---@field message string the message attached to the log
----@field webhook string url of the webhook this log should send to
+---@field webhook string shortname of the webhook this log should send to. looks up via config
```
Manason, line 104: fix back to url

modules/logger.lua
solareon: Yep as discussed if you don't configure oxlib logging it just no-ops

config/server.lua
```diff
@@ -114,6 +114,18 @@ return {
...
+        webhook = {
+            ['default'] = nil, -- default
+            ['joinleave'] = nil, -- default
+            ['ooc'] = nil, -- default
+            ['anticheat'] = nil, -- default
+            ['playermoney'] = nil, -- default
+        },
+        role = {} -- Role to tag for high priority logs. Roles use <@%roleid> and users/channels are <@userid/channelid>
+    },
+
```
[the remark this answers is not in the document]
solareon: Moved to other PR

modules/logger.lua
solareon: Removed

modules/logger.lua
```diff
@@ -103,10 +103,11 @@ end
 ---@field message string the message attached to the log
 ---@field webhook string url of the webhook this log should send to
 ---@field color? string what color the message should be
----@field tags? string[] tags in discord. Example: ['@admin', '@everyone']
+---@field tags? string[] tags in discord. Example: {'<@%roleid>', '@everyone'}
+---@field embed table formatted embed table for discord webhook
```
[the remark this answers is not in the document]
solareon: It clears an undefined field on line 76. I should probably just move the class definition up to like 45 or so there as that's where it is used.

Manason (approved):

MafewTM (approved):

FjamZoo (approved):

PR #285, opened 2023-12-17
feat(queue): add timeouts

D4isDAVID:
## Description

Adds timeouts to the queue system, letting players reconnect after a configurable amount of time after disconnecting.

Players that passed the queue and disconnected while downloading server data will also be able to reconnect after a configurable amount of time.

See #279 for additional context.

## Checklist


MafewTM (approved):

config/queue.lua
```diff
@@ -0,0 +1,8 @@
+return {
+    ---Amount of seconds to wait before removing a player from the queue after disconnecting while waiting.
+    timeoutSeconds = 30,
+
+    ---Amount of seconds to wait before removing a player from the queue after disconnecting while installing server data.
+    ---Notice that an additional ~2 minutes will be waited due to limitations with how FiveM handles joining players.
+    joiningTimeoutSeconds = 0,
```
Manason, line 7: I'm a bit confused by the nuance difference here. It seems like the first one is the timeout while the player is in the queue and the second is the timeout when the player is effectively first in line -- that is there is an open slot and they are downloading the assets to be let in? Is it worth specifying both of these config options? Why would someone want to set these to different values?

D4isDAVID: The main reason I added this is that connectqueue also has it ([link](https://github.com/Nick78111/ConnectQueue/blob/2ae29bac5ae11d549fbb319d669e9cccc2331dda/connectqueue/server/sv_queue_config.lua#L21-L25)), but connectqueue waits until the client sends an custom event to the server - meaning after the player downloaded the server content *and* passed the loading screen, however this system waits for `playerJoining`, so I don't account for players that disconnect in the loading screen.

Currently the way the queue works is that after the player is first in line, when there is an available slot, the player is let into the server and *dequeued*, so now they are installing server content and technically they are no longer first in line; they're already past that.

Obviously this can change if need be.

PR #292, opened 2023-12-26
Update LICENSE

Mesrine67:


ChatDisabled:
It is meant like that since qb-core used a lot of code from ESX. As per discussions to make QBCore "legal" one of the agreements was that qb-core would use the ESX license and give appropriate credits to ESX. Which Kakarot never did until we did it. Although most code from ESX has been removed or rewritten entirely already, I still like to give credit where it's due.

PR #295, opened 2024-01-01
Adding category to the items

bicatcho11:
Adding category to the items so that the outputs are better and more accurate
Much more than it is
Here's a real-life example

```
        for k, v in pairs(QBCore.Shared.Items) do
            if v["category"] == "ammoitems" then
                slots = slots + 1
                table.insert(tempTable, v["name"])
            end
        end
        print(json.encode(tempTable))

```

## Checklist


TonybynMp4:
Useless af, this file isn't even used after ox_inventory imported the items

Manason:
items.lua is deprecated. We don't even have a way to access it except through the bridge layer for backwards compatibility. Use ox_inventory is the source of truth for item data.

PR #298, opened 2024-01-03
feat: jobs sql table

solareon:
## Description

Creates two tables to store job information such as pay, grades, and job names. This is the first step to enable dynamic pay configuration via core exports. Currently jobs are stored in a table of tables in shared/jobs.lua. The goal is to eventually rely on that only for initial load and instead store jobs information in the database.

## Checklist


ddoucethollett:
I believe BOOLEAN type isn't supported in all mysql/mariadb versions, would be better to use TINYINT(1) as it gets casted to true/false with oxmysql anyway

Manason:
> I believe BOOLEAN type isn't supported in all mysql/mariadb versions, would be better to use TINYINT(1) as it gets casted to true/false with oxmysql anyway

Which version is it not supported in? According to the docs it seems likes it's existed for 13 years https://mariadb.com/kb/en/boolean/

PR #302, opened 2024-01-05
modify: allow for basic character creation without ox_inventory

JoeSzymkowiczFiveM:
## Description

This will allow for basic character creation to proceed even if ox_inventory is missing from the server.

## Checklist

### DIFF
--- a/server/character.lua
+++ b/server/character.lua
@@ -10,6 +10,9 @@ end
 ---@param source Source
 local function giveStarterItems(source)
     if GetResourceState('ox_inventory') == 'missing' then return end
+    while not exports.ox_inventory:GetInventory(source) do
+        Wait(100)
+    end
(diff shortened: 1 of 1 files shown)


server/character.lua
```diff
@@ -9,10 +9,7 @@ end
 
 ---@param source Source
 local function giveStarterItems(source)
-    while not exports.ox_inventory:GetInventory(source) do
```
Manason, line 12: Don't we still need this loop? This awaits for the inventory to be loaded

Manason (changes requested):

server/character.lua
JoeSzymkowiczFiveM: Good point.

Manason (approved):

MafewTM (approved):

PR #303, opened 2024-01-08
Selection of job by the user

evanillaa99:
## Description

jobByUser: Specifies whether the job is enabled for the user (true) or not (false).
He will be more flexible in dealing with, for example
qbx_cityhall

```
for k, v in pairs(THE_TABLE_NAME) do
    if v.jobByUser then
        if QBX and QBX.PlayerData and v.job then
            hasJob = false
            local playerJob = QBX.PlayerData.job.name
            for _, job in ipairs(v.job) do
                if playerJob == job then
                    hasJob = true
                    break
                end
            end
        end
    end
end
```

## Checklist

### DIFF
--- a/shared/jobs.lua
+++ b/shared/jobs.lua
@@ -171,6 +171,7 @@ return {
 		label = 'Taxi',
 		defaultDuty = true,
 		offDutyPay = false,
+        jobByUser = true,
 		grades = {
             [0] = {
                 name = 'Recruit',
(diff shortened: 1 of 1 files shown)


evanillaa99:
We may receive your comments on how to produce a more compatible form

evanillaa99:
this is my first pull :)))))))))))))))))))))))))))))))))))))))))))))))

MafewTM:
I don't think this is really necessary nor how we should do a multijob within qbx_core

evanillaa99:
I will do something better than this
In another pull

PR #307, opened 2024-01-09
Simple update for new metadata

bicatcho11:
## Description

lspd meta:
The item can be used when the server has more than one police station To extract vehicle
Evidence
The option will be faster when using the condition For example, if the vehicle is to be retrieved at the main police station, the vehicle number one is used livery 1, otherwise the vehicle police station is used livery 2, and so on

jailclothes 
To return the player's clothes after leaving prison

email for metadata
Most phone scripts now use email input
The value must be unified in the core

alcohol 

To make store alcohol values

## Checklist

### DIFF
--- a/server/player.lua
+++ b/server/player.lua
@@ -89,6 +89,7 @@ function CheckPlayerData(source, playerData)
     playerData.charinfo.gender = playerData.charinfo.gender or 0
     playerData.charinfo.backstory = playerData.charinfo.backstory or 'placeholder backstory'
     playerData.charinfo.nationality = playerData.charinfo.nationality or 'USA'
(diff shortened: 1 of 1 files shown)


Manason:
metadata values can be set from concerned resources using the set metadata function. I don't see a need to add default metadata keys for all players here.

bicatcho11:
Give me an example of this?
@Manason

bicatcho11:
What's wrong with storing alcohol values?

solareon:
What he means is that you can simply call the function https://docs.qbox.re/resources/core/player#setmetadata and it handles storing the metadata automatically. There is no need to modify the core files to add additional metadata like you had to do with qb.

PR #315, opened 2024-01-11
vehicletext v6.9

TonybynMp4:
## Description

The best version of itself

## Checklist

- [x]  My pull request fits the contribution guidelines & code conventions.

### DIFF
--- a/client/main.lua
+++ b/client/main.lua
@@ -68,7 +68,7 @@ CreateThread(function()
     for _, v in pairs(GetVehiclesByName()) do
         if v.model and v.name then
             local gameName = GetDisplayNameFromVehicleModel(v.model)
-            if gameName or gameName ~= "CARNOTFOUND" then
+            if gameName and gameName ~= "CARNOTFOUND" then
                 AddTextEntryByHash(joaat(gameName), v.name)
             end
 		end
--- a/client/main.lua
+++ b/client/main.lua
@@ -70,7 +70,9 @@ CreateThread(function()
             local gameName = GetDisplayNameFromVehicleModel(v.model)
             if gameName and gameName ~= "CARNOTFOUND" then
                 AddTextEntryByHash(joaat(gameName), v.name)
+            else
+                lib.print.warn('Could not find gameName value in vehicles.meta for vehicle model %s', v.model))
             end
-		end
+        end
 	end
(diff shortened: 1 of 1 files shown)


client/main.lua
```diff
@@ -66,8 +66,11 @@ AddTextEntry('FE_THDR_GTAO', mapText)
 
 CreateThread(function()
     for _, v in pairs(GetVehiclesByName()) do
-        if v.hash and v.name then
-			AddTextEntryByHash(v.hash, v.name)
+        if v.model and v.name then
+            local gameName = GetDisplayNameFromVehicleModel(v.model)
+            if gameName or gameName ~= "CARNOTFOUND" then
```
Manason, line 71:
```suggestion
            if gameName and gameName ~= "CARNOTFOUND" then
```
I'm assuming this is supposed to be an and. It doesn't make sense otherwise

Manason (changes requested):

client/main.lua
TonybynMp4: woopsies, i was very tired 😅

TonybynMp4:
Added a warning for missing gameNames, as it is client side it shouldn't be problematic.

client/main.lua
```diff
@@ -66,8 +66,13 @@ AddTextEntry('FE_THDR_GTAO', mapText)
...
     for _, v in pairs(GetVehiclesByName()) do
-        if v.hash and v.name then
-			AddTextEntryByHash(v.hash, v.name)
-		end
+        if v.model and v.name then
+            local gameName = GetDisplayNameFromVehicleModel(v.model)
+            if gameName and gameName ~= "CARNOTFOUND" then
+                AddTextEntryByHash(joaat(gameName), v.name)
+            else
+                lib.print.warn('Could not find gameName value in vehicles.meta for vehicle model %s', v.model)
```
Manason, line 74: Maybe instead, you should collect the models into a table and do one print?

Manason (changes requested):

client/main.lua
TonybynMp4: looks way better this way imo 🤷🏻

Manason (approved):

solareon (approved):

PR #318, opened 2024-01-15
feat: switch from utils to lib

MafewTM:
## Description

Switch from the deprecated utils module to the new lib module

## Checklist


config/server.lua
```diff
@@ -25,7 +25,7 @@ return {
         identifierTypes = {
             citizenid = {
                 valueFunction = function()
-                    return tostring(RandomLetter(3) .. RandomNumber(5)):upper()
+                    return tostring(lib.string.random('A', 3) .. lib.string.random(1, 2))
```
Manason, line 28:
```suggestion
                    return tostring(lib.string.random('A', 3) .. lib.string.random('1', 2))
```
ox docs say that the first param is a string. Also, isn't this a change in behavior? The previous value function was five random numbers and now this is two?

config/server.lua
```diff
@@ -40,7 +40,7 @@ return {
             },
             FingerId = {
                 valueFunction = function()
-                    return tostring(RandomLetter(2) .. RandomNumber(3) .. RandomLetter(1) .. RandomNumber(2) .. RandomLetter(3) .. RandomNumber(4))
+                    return tostring(lib.string.random(1, 2) .. lib.string.random(1, 3) .. lib.string.random(1, 1) .. lib.string.random(1, 2) .. lib.string.random(1, 3) .. lib.string.random(1, 4))
```
Manason, line 43: ditto

Manason (changes requested):

Manason (approved):

server/commands.lua
```diff
@@ -142,8 +142,15 @@ lib.addCommand('car', {
...
     if not args then return end
-    local netId = SpawnVehicle(source, args[Lang:t("command.car.params.model.name")], nil, true)
-    local plate = GetPlate(NetworkGetEntityFromNetworkId(netId))
+    local ped = GetPlayerPed(source)
+
+    local netId = qbx.spawnVehicle({
+        model = args[Lang:t("command.car.params.model.name")],
+        spawnSource = ped,
+        warp = ped,
+    })
```
D4isDAVID, line 151:
```suggestion
    local netId = qbx.spawnVehicle({
        model = args[Lang:t("command.car.params.model.name")],
        spawnSource = GetPlayerPed(source),
        warp = true,
    })
```
It's possible to do this instead

fxmanifest.lua
```diff
@@ -7,14 +7,14 @@ version '1.4.0'
...
-    'modules/utils.lua',
+    'modules/lib.lua',
     'shared/locale.lua',
     'locale/en.lua',
     'locale/*.lua',
 }
 
 client_scripts {
-    'modules/utils.lua',
+    'modules/lib.lua',
```
D4isDAVID, line 17: Do we need this extra declaration? This already exists in `shared_scripts`.

bridge/qb/client/functions.lua
```diff
@@ -86,56 +121,97 @@ end
...
+end
 
----@deprecated use GetPlayersInScope from imports/utils.lua
+---@deprecated use the GetPlayersInScope native directly
 functions.GetPlayers = GetPlayersInScope
 
----@deprecated use GetPeds from imports/utils.lua
-functions.GetPeds = GetPeds
+---@deprecated use the GetGamePool('CPed') native directly
+functions.GetPeds = function()
```
D4isDAVID, line 138: The original [`QBCore.Functions.GetPeds`](https://github.com/qbcore-framework/qb-core/blob/cdd1502f0157fe2c9e1d97ef1b6e79eceb9e6c9f/client/functions.lua#L169) actually supports an `ignoreList` that was implemented in utils. We should probably re-implement this inside the bridge.

bridge/qb/client/functions.lua
```diff
@@ -86,56 +121,97 @@ end
...
+    GetGamePool('CPed')
+end
 
----@deprecated use GetClosestPed from imports/utils.lua
+---@deprecated use lib.getClosestPed from ox_lib
 ---Use GetClosestPlayer if wanting to ignore non-player peds
-functions.GetClosestPed = GetClosestPed
+functions.GetClosestPed = lib.getClosestPed
```
D4isDAVID, line 144: Ditto, also lib returns different types of results than the original [`QBCore.Functions.GetClosestPed`](https://github.com/qbcore-framework/qb-core/blob/cdd1502f0157fe2c9e1d97ef1b6e79eceb9e6c9f/client/functions.lua#L187) function.

bridge/qb/client/functions.lua
```diff
@@ -86,56 +121,97 @@ end
...
 
----@deprecated use IsWearingGloves from imports/utils.lua
-functions.IsWearingGloves = IsWearingGloves
+---@deprecated use qbx.isWearingGloves from modules/lib.lua
+functions.IsWearingGloves = qbx.isWearingGloves
 
----@deprecated use GetClosestPlayer from imports/utils.lua
-functions.GetClosestPlayer = GetClosestPlayer
+---@deprecated use lib.getClosestPlayer from ox_lib
+functions.GetClosestPlayer = lib.getClosestPlayer
```
D4isDAVID, line 150: This one also returns different results.

bridge/qb/client/functions.lua
```diff
@@ -86,56 +121,97 @@ end
...
 
----@deprecated use GetPlayersFromCoords from imports/utils.lua
-functions.GetPlayersFromCoords = GetPlayersFromCoords
+---@deprecated use lib.getNearbyPlayers from ox_lib
+functions.GetPlayersFromCoords = lib.getNearbyPlayers
```
D4isDAVID, line 153: Ditto.

bridge/qb/client/functions.lua
```diff
@@ -86,56 +121,97 @@ end
...
 
----@deprecated use GetClosestVehicle from imports/utils.lua
-functions.GetClosestVehicle = GetClosestVehicle
+---@deprecated use lib.getClosestVehicle from ox_lib
+functions.GetClosestVehicle = lib.getClosestVehicle
```
D4isDAVID, line 156: Ditto.

bridge/qb/client/functions.lua
```diff
@@ -86,56 +121,97 @@ end
...
 
----@deprecated use GetClosestObject from imports/utils.lua
-functions.GetClosestObject = GetClosestObject
+---@deprecated use lib.getClosestObject from ox_lib
+functions.GetClosestObject = lib.getClosestObject
```
D4isDAVID, line 159: Ditto.

bridge/qb/client/functions.lua
```diff
@@ -153,17 +229,17 @@ function functions.SpawnVehicle(model, cb, coords, isnetworked, teleportInto)
...
 
----@deprecated use DeleteVehicle from imports/utils.lua
-functions.DeleteVehicle = DeleteVehicle
+---@deprecated use qbx.deleteVehicle from modules/lib.lua
+functions.DeleteVehicle = qbx.deleteVehicle
 
----@deprecated use GetPlate from imports/utils.lua
-functions.GetPlate = GetPlate
+---@deprecated use qbx.getVehiclePlate from modules/lib.lua
+functions.GetPlate = qbx.getVehiclePlate
```
D4isDAVID, line 236: [`QBCore.Functions.GetPlate`](https://github.com/qbcore-framework/qb-core/blob/cdd1502f0157fe2c9e1d97ef1b6e79eceb9e6c9f/client/functions.lua#L386) handles the `vehicle` param differently in some cases. Maybe we should too to improve compatibility?

bridge/qb/client/functions.lua
```diff
@@ -153,17 +229,17 @@ function functions.SpawnVehicle(model, cb, coords, isnetworked, teleportInto)
...
 
----@deprecated use GetVehicleDisplayName from imports/utils.lua
-functions.GetVehicleLabel = GetVehicleDisplayName
+---@deprecated use qbx.getVehicleDisplayName from modules/lib.lua
+functions.GetVehicleLabel = qbx.getVehicleDisplayName
```
D4isDAVID, line 239: [Ditto.](https://github.com/qbcore-framework/qb-core/blob/cdd1502f0157fe2c9e1d97ef1b6e79eceb9e6c9f/client/functions.lua#L391)

bridge/qb/client/functions.lua
```diff
@@ -153,17 +229,17 @@ function functions.SpawnVehicle(model, cb, coords, isnetworked, teleportInto)
...
 
----@deprecated use IsVehicleSpawnClear from imports/utils.lua
-functions.SpawnClear = IsVehicleSpawnClear
+---@deprecated use lib.getNearbyVehicles from ox_lib
+functions.SpawnClear = lib.getNearbyVehicles
```
D4isDAVID, line 242: These ones have completely different return types.

bridge/qb/client/functions.lua
```diff
@@ -621,26 +696,117 @@ end
...
+end
 
----@deprecated use GetStreetNameAtCoords from imports/utils.lua
-functions.GetStreetNametAtCoords = GetStreetNameAtCoords
+---@deprecated use qbx.getStreetName from modules/lib.lua
+functions.GetStreetNametAtCoords = qbx.getStreetName
 
----@deprecated use GetZoneAtCoords from imports/utils.lua
-functions.GetZoneAtCoords = GetZoneAtCoords
+---@deprecated  use qbx.getZoneName from modules/lib.lua
```
D4isDAVID, line 772:
```suggestion
---@deprecated use qbx.getZoneName from modules/lib.lua
```
Just a nitpick for a single whitespace character.

bridge/qb/client/functions.lua
```diff
@@ -621,26 +696,117 @@ end
...
+functions.GetZoneAtCoords = qbx.getZoneName
 
----@deprecated use GetCardinalDirection from imports/utils.lua
-functions.GetCardinalDirection = GetCardinalDirection
+---@deprecated use qbx.getCardinalDirection from modules/lib.lua
+functions.GetCardinalDirection = qbx.getCardinalDirection
```
D4isDAVID, line 776: [`QBCore.Functions.GetCardinalDirection`](https://github.com/qbcore-framework/qb-core/blob/cdd1502f0157fe2c9e1d97ef1b6e79eceb9e6c9f/client/functions.lua#L965) returns a `Cardinal Direction Error` for entities that don't exist. Maybe we should do that too?

bridge/qb/server/main.lua
```diff
@@ -85,15 +85,31 @@ function qbCoreCompat.Functions.TriggerCallback(name, source, cb, ...)
...
 qbCoreCompat.Functions.CreateCallback('QBCore:Server:SpawnVehicle', function(source, cb, model, coords, warp)
-    local netId = SpawnVehicle(source, model, coords, warp)
+    model = type(model) == 'string' and joaat(model) or (model --[[@as integer]])
+    local ped = GetPlayerPed(source)
+
+    local netId = qbx.spawnVehicle({
+        model = model,
+        spawnSource = coords or ped,
+        warp = warp and ped or nil,
+    })
```
D4isDAVID, line 97:
```suggestion
    local netId = qbCoreCompat.Functions.SpawnVehicle(source, model, coords, warp)
```
Actually I believe you can just do this

bridge/qb/server/main.lua
```diff
@@ -85,15 +85,31 @@ function qbCoreCompat.Functions.TriggerCallback(name, source, cb, ...)
...
 qbCoreCompat.Functions.CreateCallback('QBCore:Server:CreateVehicle', function(source, cb, model, coords, warp)
-    local netId = SpawnVehicle(source, model, coords, warp)
+    model = type(model) == 'string' and joaat(model) or (model --[[@as integer]])
+    local ped = GetPlayerPed(source)
+
+    local netId = qbx.spawnVehicle({
+        model = model,
+        spawnSource = coords or ped,
+        warp = warp and ped or nil,
+    })
```
D4isDAVID, line 111:
```suggestion
    local netId = qbCoreCompat.Functions.CreateVehicle(source, model, nil, coords, warp)
```
Same here but a kinda different function.

bridge/qb/shared/main.lua
```diff
@@ -1,36 +1,43 @@
...
+---@deprecated use lib.string.random from ox_lib
+qbShared.RandomInt = function(length)
+    if length <= 0 then return '' end
+    return qbShared.RandomInt(length - 1) .. lib.string.random('1')
+end
 
----@deprecated use string.split from imports/utils.lua
-qbShared.SplitStr = string.split
+---@deprecated use string.strsplit with CfxLua 5.4
+qbShared.SplitStr = string.strsplit
```
D4isDAVID, line 20: These return different result types.

bridge/qb/shared/main.lua
```diff
@@ -1,36 +1,43 @@
...
 
----@deprecated use string.trim from imports/utils.lua
-qbShared.Trim = string.trim
+---@deprecated use qbx.string.trim from modules/lib.lua
+qbShared.Trim = qbx.string.trim
```
D4isDAVID, line 23: Once again there is a case here that [`QBShared.Trim`](https://github.com/qbcore-framework/qb-core/blob/cdd1502f0157fe2c9e1d97ef1b6e79eceb9e6c9f/shared/main.lua#L40) handles but we don't anymore.

bridge/qb/shared/main.lua
```diff
@@ -1,36 +1,43 @@
...
 
----@deprecated use string.firstToUpper from imports/utils.lua
-qbShared.FirstToUpper = string.firstToUpper
+---@deprecated use qbx.string.capitalize from modules/lib.lua
+qbShared.FirstToUpper = qbx.string.capitalize
```
D4isDAVID, line 26: [Ditto.](https://github.com/qbcore-framework/qb-core/blob/cdd1502f0157fe2c9e1d97ef1b6e79eceb9e6c9f/shared/main.lua#L45)

config/server.lua
```diff
@@ -25,7 +25,7 @@ return {
         identifierTypes = {
             citizenid = {
                 valueFunction = function()
-                    return tostring(RandomLetter(3) .. RandomNumber(5)):upper()
+                    return tostring(lib.string.random('A', 3) .. lib.string.random('1', 5))
```
D4isDAVID, line 28:
```suggestion
                    return lib.string.random('........')
```
This isn't how this function works. Also why don't we just create 8 random alphanumerical characters instead of 3 letters and 5 numbers?

config/server.lua
```diff
@@ -40,7 +40,7 @@ return {
             },
             FingerId = {
                 valueFunction = function()
-                    return tostring(RandomLetter(2) .. RandomNumber(3) .. RandomLetter(1) .. RandomNumber(2) .. RandomLetter(3) .. RandomNumber(4))
+                    return tostring(lib.string.random('1', 2) .. lib.string.random('1', 3) .. lib.string.random('1', 1) .. lib.string.random('1', 2) .. lib.string.random('1', 3) .. lib.string.random('1', 4))
```
D4isDAVID, line 43:
```suggestion
                    return lib.string.random('...............')
```
Ditto.

D4isDAVID (changes requested):
There are some comments here about cases that QBCore handles but we don't. I'm not exactly strict about this, but I believe it would be nice to handle those cases ourselves too to improve compatibility.

There are also some ped-related functions that support an `ignoreList` in QBCore but not in lib anymore, it's possible to copy some stuff from the utils module to do that.

bridge/qb/client/functions.lua
MafewTM: Do we want it to just return end if there is no plate? I'd think it erroring out would be better than just doing that

bridge/qb/client/functions.lua
```diff
@@ -86,56 +121,111 @@ end
...
 
----@deprecated use GetObjects from imports/utils.lua
-functions.GetObjects = GetObjects
+---@deprecated use the GetGamePool('CObject') native directly
+functions.GetObjects = function()
+    GetGamePool('CObject')
+end
 
----@deprecated use GetPlayersInScope from imports/utils.lua
+---@deprecated use the GetPlayersInScope native directly
```
BerkieBb, line 134: There is no GetPlayersInScope native

bridge/qb/client/functions.lua
BerkieBb: If we aren't, that should be in there definitely

bridge/qb/client/functions.lua
MafewTM: Whats the plan for this one @D4isDAVID

D4isDAVID: I think I meant to write `GetActivePlayers` which this function uses internally.

bridge/qb/client/functions.lua
MafewTM: Everything else is pretty much done except for this function. Trying to Frankenstein this together is kinda confusing so help on this would be appreciated. Take another pass through as well and make sure I did everything right

BerkieBb: The one from utils correctly converts it

BerkieBb:
All comments resolved now

modules/lib.lua
```diff
@@ -90,6 +173,10 @@ end
 ---@param entity integer
 ---@return 'North' | 'South' | 'East' | 'West'
 function qbx.getCardinalDirection(entity)
+    if not DoesEntityExist(entity) then
+        return lib.print.error('Cardinal Direction Error')
+    end
+
```
D4isDAVID, line 179: I'm not sure whether I want this added directly to the lib or just handled by utils & the QB bridge. I kinda didn't want to handle user errors in the lib, much less like this because it adds a weird hardcoded return value.

bridge/qb/client/functions.lua
```diff
@@ -153,17 +266,27 @@ function functions.SpawnVehicle(model, cb, coords, isnetworked, teleportInto)
...
+    return qbx.getVehicleDisplayName(vehicle)
+end
 
----@deprecated use IsVehicleSpawnClear from imports/utils.lua
-functions.SpawnClear = IsVehicleSpawnClear
+---@deprecated use lib.getNearbyVehicles from ox_lib
+functions.SpawnClear = function(coords, radius)
+    local vehicles = lib.getNearbyVehicles(coords, radius)
+    if #vehicles > 0 then return false end
+    return true
```
D4isDAVID, line 288:
```suggestion
    return #vehicles == 0
```
This can be simplified.

bridge/qb/client/functions.lua
```diff
@@ -86,56 +121,134 @@ end
...
 ---@param model string|number
 ---@param cb? fun(vehicle: number)
 ---@param coords? vector4 player position if not specified
 ---@param isnetworked? boolean defaults to true
 ---@param teleportInto boolean teleport player to driver seat if true
 function functions.SpawnVehicle(model, cb, coords, isnetworked, teleportInto)
-    coords = type(coords) == 'table' and vec4(coords.x, coords.y, coords.z, coords.w or GetEntityHeading(cache.ped)) or coords or GetCoordsFromEntity(cache.ped)
+    local playerCoords = GetEntityCoords(cache.ped)
+    local combinedCoords = vec4(playerCoords.x, playerCoords.y, playerCoords.z, GetEntityHeading(cache.ped))
+    coords = type(coords) == 'table' and vec4(coords.x, coords.y, coords.z, coords.w or GetEntityHeading(cache.ped)) or coords or combinedCoords
```
D4isDAVID, line 251:
```suggestion
    coords = type(coords) == 'table' and vec4(coords.x, coords.y, coords.z, coords.w or combinedCoords.w) or coords or combinedCoords
```
I guess we can re-use the heading instead of calculating it again.

D4isDAVID (changes requested):

modules/lib.lua
BerkieBb: lib.print.error doesn't return anything either so, this is still correct

D4isDAVID: It isn't correct both in the sense that the function would return `nil` when I don't want the function to handle that, and also in the sense that this makes the utils & the QB bridge not actually handle this the same way QB does (which is why this was added in the first place).

BerkieBb: Either we change it here to just return the string or we change the bridge to return that when it returns nil

D4isDAVID: I want this to not be handled by the lib at all, and instead to have utils and the QB bridge wrap this function and check for the entity beforehand.

Other string and vehicle related functions had this type of behavior removed so I don't want this  added here either.

BerkieBb: The utils will be removed soon, so I only added it to the bridge

D4isDAVID: I forgot that I already did the utils part, so yeah only the bridge was needed.

D4isDAVID (approved):
Looks good. I guess the only thing left is some manual testing.

PR #328, opened 2024-01-22
Transferring the command from the qbx_phone to qbx_core

bicatcho11:
## Description

Transferring the command from the qbx_phone to qbx_core

I think he has his rightful place here, and the matter can develop to deal with all metadata

After approval, the matter will be developed to deal with all cases
addition metadata
delete metadata

## Checklist

### DIFF
--- a/server/commands.lua
+++ b/server/commands.lua
@@ -337,6 +337,21 @@ lib.addCommand('me', {
     playerState:set('me', nil, true)
 end)
 
+-- Metadata command
+
+lib.addCommand('setmetadata', {help = 'Set Player Metadata (God Only)', restricted = 'god'}, function(source, args)
+    local Player = exports.qbx_core:GetPlayer(source)
+    if args[1] then
+        if args[1] == "trucker" then
+            if args[2] then
+                local newrep = Player.PlayerData.metadata["jobrep"]
+                newrep.trucker = tonumber(args[2])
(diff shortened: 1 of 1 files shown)


solareon:
This seems more like something that should go into https://github.com/Qbox-project/qbx_adminmenu

server/commands.lua
```diff
@@ -337,6 +337,21 @@ lib.addCommand('me', {
...
+    if args[1] then
+        if args[1] == "trucker" then
+            if args[2] then
+                local newrep = Player.PlayerData.metadata["jobrep"]
+                newrep.trucker = tonumber(args[2])
+                Player.Functions.SetMetaData("jobrep", newrep)
+            end
+        end
+    end
+end)
```
Manason, line 353: The body doesn't seem to match the name. I read the name and thought this was a general way for admins to set metadata on a player (could be dangerous as admins could easily break contracts by saving data in the wrong format), but then I read the body and the command actually just seems to set a very specific metadata value "jobrep.trucker"

I'm not sure why this would belong in core? It seems very specific to a trucker resource.

Manason (changes requested):

server/commands.lua
bicatcho11: I will keep it open to all metadata not only trucker

bicatcho11:
I will close this pull, And open a new one on Thursday with the new order
I want to try some experiments

PR #337, opened 2024-01-28
fix(bridge/qb/client/functions): restore backwards compatibility

Manason:
v1.5.0 broke backwards compatibility by changing some function headers and behaviors. This restores the behavior to what it was prior to v1.5.0

### DIFF
--- a/bridge/qb/client/functions.lua
+++ b/bridge/qb/client/functions.lua
@@ -215,7 +215,7 @@ end
 ---@deprecated use lib.getNearbyPlayers from ox_lib
 functions.GetPlayersFromCoords = function(coords, radius)
     coords = type(coords) == 'table' and vec3(coords.x, coords.y, coords.z) or coords or GetEntityCoords(cache.ped)
-    local players = lib.getNearbyPlayers(coords, distance or 5, true)
+    local players = lib.getNearbyPlayers(coords, radius or 5, true)
 
     -- This is for backwards compatability as beforehand it only returned the PlayerId, where Lib returns PlayerPed, PlayerId and PlayerCoords
     for i = 1, #players do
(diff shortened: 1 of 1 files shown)


D4isDAVID:
The original QBCore functions `GetVehicles`, `GetObjects`, `GetPlayers`, `GetClosestVehicle` and `GetClosestObject` don't actually take in an `ignoreList`. The original `GetClosestPlayer` doesn't take a `maxDistance` either.

Although it was like this before v1.5.0, is this not invalid? Or is this already used by some users?

BerkieBb (approved):

bridge/qb/client/functions.lua
```diff
@@ -285,7 +323,18 @@ end
...
+    local vehicles = GetGamePool('CVehicle')
+    local closeVeh = {}
+    for i = 1, #vehicles do
+        local vehicleCoords = GetEntityCoords(vehicles[i])
+        local distance = #(vehicleCoords - coords)
+        if distance <= radius then
+            closeVeh[#closeVeh + 1] = vehicles[i]
+        end
+    end
+    return #closeVeh == 0
```
D4isDAVID, line 337: Why can't we keep using `lib.getNearbyVehicles` here?

bridge/qb/client/functions.lua
```diff
@@ -121,80 +121,118 @@ end
...
-        if not ignoreMap[entity] then
-            peds[#peds + 1] = entity
+    for i = 1, #plys do
+        local player = plys[i]
+        if not ignoreMap[player] then
+            players[#players + 1] = player
         end
     end
-    return peds
+    return players
```
D4isDAVID, line 171: Suggestion: it is possible to refactor `getEntities` to accept a pool of entities or a list of players instead of a pool name to use it here too.

D4isDAVID (changes requested):

D4isDAVID (approved):

PR #344, opened 2024-02-04
feat: Add disableDrawRect param

TonybynMp4:
## Description

Lets you disable the drawRect which can be really ugly & not that necessary in some cases
## Checklist

### DIFF
--- a/modules/lib.lua
+++ b/modules/lib.lua
@@ -294,7 +294,7 @@ else
 
     ---@class LibDrawText3DParams : LibDrawTextParams
     ---@field coords vector3
-    ---@field drawRect? boolean default: `true`
+    ---@field disableDrawRect? boolean default: `true`
 
     ---Draws text onto the screen in 3D space for a single frame.
     ---@param params LibDrawText3DParams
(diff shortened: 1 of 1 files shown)


modules/lib.lua
```diff
@@ -294,6 +294,7 @@ else
 
     ---@class LibDrawText3DParams : LibDrawTextParams
     ---@field coords vector3
+    ---@field drawRect? boolean default: `true`
```
D4isDAVID, line 297: Maybe make it `disableDrawRect` instead?

TonybynMp4: yeah maybe that'd be better lol

modules/lib.lua
```diff
@@ -294,6 +294,7 @@ else
 
     ---@class LibDrawText3DParams : LibDrawTextParams
     ---@field coords vector3
+    ---@field disableDrawRect? boolean default: `true`
```
D4isDAVID, line 297:
```suggestion
    ---@field disableDrawRect? boolean
```
This can be removed now

D4isDAVID (changes requested):

D4isDAVID (approved):

Manason (approved):

PR #345, opened 2024-02-04
fix: send player source to server event

TonybynMp4:
## Description

might be interesting to know who's metadata has been changed 🙄 
## Checklist

### DIFF
--- a/server/player.lua
+++ b/server/player.lua
@@ -324,7 +324,7 @@ function CreatePlayer(playerData, Offline)
             self.Functions.Save()
         end
         TriggerClientEvent('qbx_core:client:onSetMetaData', self.PlayerData.source, meta, oldVal, val)
-        TriggerEvent('qbx_core:server:onSetMetaData', self.PlayerData.source, meta,  oldVal, val)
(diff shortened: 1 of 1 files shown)


server/player.lua
```diff
@@ -324,7 +324,7 @@ function CreatePlayer(playerData, Offline)
             self.Functions.Save()
         end
         TriggerClientEvent('qbx_core:client:onSetMetaData', self.PlayerData.source, meta, oldVal, val)
-        TriggerEvent('qbx_core:server:onSetMetaData', meta,  oldVal, val)
+        TriggerEvent('qbx_core:server:onSetMetaData', self.PlayerData.source, meta,  oldVal, val)
```
Manason, line 327: This is a breaking change. Please put source as the last value.

Manason (changes requested):

server/player.lua
TonybynMp4: does it really matter? it's so recent that there's probably no one using it yet 🤔

solareon (approved):

D4isDAVID (approved):

Manason (approved):

PR #346, opened 2024-02-05
add disableControls function easy way

bicatcho11:
## Description

Definition of the function qbx.disableControls: This function is used to disable game controls except for the specified commands.

Creation of the controlActions table: This table contains the commands that need to be disabled, along with their associated keys.

Creation of a loop: This loop is where the commands are disabled based on the input values, which are taken from the controlActions table.

qbx.disableControls({
            enableMouse = false,
            enableMovement = true,
            enableCarMovement = true,
            enableCombat = false,
            enablePause = true
        })

## Checklist

### DIFF
--- a/modules/lib.lua
+++ b/modules/lib.lua
@@ -4,6 +4,7 @@ local qbx = {}
 qbx.string = {}
 qbx.math = {}
 qbx.table = {}
+controlsLoop = nil
 
 qbx.armsWithoutGloves = lib.table.freeze({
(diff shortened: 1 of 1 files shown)


tom-osborne:
This functionality is already provided by ox_lib. Not sure it is necessary to add this to the core.

https://overextended.dev/ox_lib/Modules/DisableControls/Client#libdisablecontrols

PR #348, opened 2024-02-08
feat(dynamic groups): sql schema

Manason:
Moves groups from config into database. Config would be the default values and the database would override.
This allows for jobs/gangs to be modified at runtime supporting features such as players changing names.
Opted for a more flexible structure since normal use cases would be reading/writing to an individual group/grade. Querying a selection of grades/groups doesn't seem to be a likely use case.

### DIFF
--- a/qbx_core.sql
+++ b/qbx_core.sql
@@ -42,3 +42,17 @@ CREATE TABLE IF NOT EXISTS `player_contacts` (
   PRIMARY KEY (`id`),
   KEY `citizenid` (`citizenid`)
 ) ENGINE=InnoDB AUTO_INCREMENT=1;
+
+CREATE TABLE IF NOT EXISTS `jobs` (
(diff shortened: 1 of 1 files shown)


TonybynMp4:
I hate this so much

qbx_core.sql
```diff
@@ -42,3 +42,19 @@ CREATE TABLE IF NOT EXISTS `player_contacts` (
   PRIMARY KEY (`id`),
   KEY `citizenid` (`citizenid`)
 ) ENGINE=InnoDB AUTO_INCREMENT=1;
+
+CREATE TABLE IF NOT EXISTS `groups` (
+	`name` VARCHAR(50) NOT NULL,
+  `type` ENUM('job','gang') NOT NULL,
```
BerkieBb, line 48: indentation

MafewTM (approved):

PR #350, opened 2024-02-09
add middle name for character

bicatcho11:
## Description

add middle name for character

## Checklist


MafewTM:
I personally don't see any real use for this other than maybe for some sort of RP purpose. If we do support this, this should be a configurable option. Server owners should have the ability to enable or disable this field.

Manason:
I think this will make things more complex then they need to be. A lot of scripts already assume the first name + last name combo, and if wanting a middle name for RP, one could add it to the list name already.

bicatcho11:
> I personally don't see any real use for this other than maybe for some sort of RP purpose. If we do support this, this should be a configurable option. Server owners should have the ability to enable or disable this field.

```        {
            type = 'input',
            required = false,
            icon = 'user-pen',
            label = locale('info.middle_name'),
            placeholder = 'Middle'
        },```
        
        Is this good ?

Manason:
This is a feature that would need greater discussion on if it's going to be done, and if so, how it would be done. Please open an issue if you want to suggest this.

PR #355, opened 2024-02-11
Update player.lua

SKITTLE6969:
I hate decimals!

## Checklist

### DIFF
--- a/server/player.lua
+++ b/server/player.lua
@@ -348,7 +348,7 @@ function CreatePlayer(playerData, Offline)
     ---@return boolean success if money was added
     function self.Functions.AddMoney(moneytype, amount, reason)
         reason = reason or 'unknown'
-        amount = tonumber(amount) --[[@as number]]
+        amount = math.floor(tonumber(amount)) --[[@as number]]
         if amount < 0 then return false end
         if not self.PlayerData.money[moneytype] then return false end
         self.PlayerData.money[moneytype] = self.PlayerData.money[moneytype] + amount
@@ -378,7 +378,7 @@ function CreatePlayer(playerData, Offline)
     ---@return boolean success if money was removed
     function self.Functions.RemoveMoney(moneytype, amount, reason)
         reason = reason or 'unknown'
-        amount = tonumber(amount) --[[@as number]]
+        amount = math.floor(tonumber(amount)) --[[@as number]]
         if amount < 0 then return false end
         if not self.PlayerData.money[moneytype] then return false end
         for _, mtype in pairs(config.money.dontAllowMinus) do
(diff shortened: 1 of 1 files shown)


Manason (approved):
Maybe this should round instead of flooring?

MafewTM (approved):
If we aren't wanting players to use decimals then sure

D4isDAVID:
> > Maybe this should round instead of flooring?
> 
> Math.round(x) is not exactly the same as Math.floor(x + 0.5) . When x is -0, or -0.5 ≤ x < 0, Math.round(x) returns -0, while Math.floor(x + 0.5) returns 0. However, neglecting that difference and potential precision errors, Math.round(x) and Math.floor(x + 0.5) are generally equivalent.
> 
> I just googled this tis up to you, i dont see a big difference tbh xD

Not sure why bring up JavaScript's documentation into this conversation. The question was whether it would be better to round the number up when the fraction is >=0.5, instead of always rounding the number down.

Manason:
I think either is fine. Has this been tested?

BerkieBb:
I don't think we should do this though, let people use decimals. If you don't want it, keep the change to yourself

Manason:
> I don't think we should do this though, let people use decimals. If you don't want it, keep the change to yourself

I don't think decimals are possible? I don't think it's supported in base GTA money, nor do I believe ox_inventory supports decimal amounts of items

BerkieBb:
> > I don't think we should do this though, let people use decimals. If you don't want it, keep the change to yourself
> 
> I don't think decimals are possible? I don't think it's supported in base GTA money, nor do I believe ox_inventory supports decimal amounts of items

We are not using a base gta system but I forgot that we're using ox inventory

solareon:
> > > I don't think we should do this though, let people use decimals. If you don't want it, keep the change to yourself
> > 
> > 
> > I don't think decimals are possible? I don't think it's supported in base GTA money, nor do I believe ox_inventory supports decimal amounts of items
> 
> /givemoney ID cash/bank 9.9999999999 and check eventhough inventory rounds the numbers, the hud dosent so the inventory will continue to remain a whole number while the cash and bank decimals will prevail

I believe that we should just be flooring this value. ox_inventory doesn't give you fractions of money nor does it support pricing items at fractions of money so carrying the decimal around is pointless.

BerkieBb:
Flooring would lead to improper values, as you'd have 5.9 and then suddenly just 5

solareon:
> Flooring would lead to improper values, as you'd have 5.9 and then suddenly just 5

Fair point. this should be using qbx.math.round as math.round isn't a thing in lua 5.4 and this was provided via the utils module.

BerkieBb (approved):

solareon (approved):

PR #356, opened 2024-02-11
refactor: move groups out of shared

Manason:
- Only bridge has QBX.Shared.Jobs and QBX.Shared.Gangs now to maintain backwards compatibility
- In non-bridge, the Shared group tables have now been moved to a private file with getters and setters
- New events are used to cache the gangs and jobs on the client
- New exports GetGang and GetJob introduced. Since the data can now change at runtime, callers of core have the option to either cache the groups locally and update via event, or call exports to get a specific group without storing the data


MafewTM (approved):

bridge/qb/shared/main.lua
```diff
@@ -50,4 +50,7 @@ qbShared.MaleNoGloves = qbx.armsWithoutGloves.male
 ---@deprecated use qbx.armsWithoutGloves.female from modules/lib.lua
 qbShared.FemaleNoGloves = qbx.armsWithoutGloves.female
 
+qbShared.Jobs = IsDuplicityVersion() and GetJobs() or lib.callback.await('qbx_core:server:getJobs')
+qbShared.Gangs = IsDuplicityVersion and GetGangs() or lib.callback.await('qbx_core:server:getGangs')
```
BerkieBb, line 54: Forgot to call function here

BerkieBb (approved):

PR #358, opened 2024-02-12
Use tax equations in the core (need test)

makaman50:
## Description

Use tax equations in the project

I will confirm the approval in the coming days to present a tax system based on the level system from the first to the tenth level The value will be deducted from the player according to the individual’s income

## Checklist

### DIFF
--- a/client/functions.lua
+++ b/client/functions.lua
@@ -1,4 +1,5 @@
 local positionConfig = require 'config.shared'.notifyPosition
+local moneyTax = require 'config.shared'.moneyTax
 
 ---Text box popup for player which dissappears after a set time.
 ---@param text table|string text of the notification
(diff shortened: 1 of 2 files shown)


Manason:
Can you create an issue for this? This would need to be discussed and I'm not sure a PR is the correct place.

BerkieBb:
This seems a little too personal of a change for our core

FjamZoo:
If there should be a tax involved it should be a global state called "tax" or whatever.

That way it can be accessible client/server for whatever people would need

makaman50:
There is nothing personal in this code, the idea is to specify the currency used on the server from a real currency and a virtual currency
For example, the dollar $ and Bitcoin
A broader tax system must be defined in this regard, especially when players have a lot of money inside the server
Taxes on cars and food could be raised and etc. and controlled through the core

BerkieBb:
This is more personal due to the fact that not everyone will use tax, this is a very specific rp related feature that I don't think belongs in our core due to the amount of other things you can use our core for.

makaman50:
For example, on the subject of currency, we have done so in the first version qbus
Convert the euro currency to dollars everywhere in the code, One currency must be specified and changed from one place into core
Also, the system of raising and determining taxes on the server allows server owners to have broad control over the server economy and reduce and increase the money inside it.

MafewTM:
I also don't believe this belongs in our core. Besides this purely being a math function with no actual tax functionality, I don't think qbox needs to have a native tax system.

Manason:
As I said before, I believe there needs to be discussion around exactly how this system would work, what it would do, etc. Please open an issue if you want to suggest this feature.

PR #359, opened 2024-02-12
feat: jobs/gangs stored in database instead of config

Manason:
- On initial load, gangs/jobs are fetched from the database. Any groups in the config, but not in the database are added to the database. It's intended that server operators edit the database directly and stop using the job/gang config. Future PRs will make this easier by adding menus and commands to edit groups with.

### DIFF
--- a/fxmanifest.lua
+++ b/fxmanifest.lua
@@ -26,6 +26,8 @@ client_scripts {
 
 server_scripts {
     '@oxmysql/lib/MySQL.lua',
+    'server/storage/players.lua',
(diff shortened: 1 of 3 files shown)


BerkieBb (approved):

solareon (approved):

Filcatalao:
What will be the advantage of this (like old esx) having it on a table? Managing existing jobs will be made on the shared jobs.lua or database? (for example to change names or salary values etc on a existing job or even add new roles, but doing it offline and not ingame)

PR #370, opened 2024-02-17
feat: multi job/group implementation

Manason:
- Introduces new playerdata jobs & gangs tables to hold key value data of all the player's groups (name & grade), including the primary. The only exception is the unemployed/none groups so that these don't contribute to the player limit.
- The existing playerdata.job/gang is the primary group, and is populated with all the data about that group and grade.
- New exports added for managing player jobs/gangs (add, remove, set primary)
- To maintain backwards compatibility, SetJob & SetGang will replace the primary job rather than act as an alias for Add.
- New config options to limit the total number of jobs/gangs a single player may hold at a time

The multi-group system is not optional and cannot be disabled. While using external multigroup systems is still possible, unless they use the exports to propagate the data to the qbox multi-group, it may break other scripts that rely on that data being accurate. I considered a way to make it optional, but doing so would have the opposite effect of making it so that other scripts couldn't rely on the built-in multi-group functionality.

### DIFF
--- a/config/server.lua
+++ b/config/server.lua
@@ -137,7 +137,4 @@ return {
     removeSocietyMoney = function(accountName, payment)
         return exports['Renewed-Banking']:removeAccountMoney(accountName, payment)
     end,
-
-    maxJobsPerPlayer = 1,
-    maxGangsPerPlayer = 1,
 }
--- a/config/server.lua
+++ b/config/server.lua
@@ -136,5 +136,8 @@ return {
 
     removeSocietyMoney = function(accountName, payment)
         return exports['Renewed-Banking']:removeAccountMoney(accountName, payment)
-    end
+    end,
+
+    maxJobsPerPlayer = 1,
+    maxGangsPerPlayer = 1,
 }
--- a/server/player.lua
+++ b/server/player.lua
@@ -4,6 +4,8 @@ local logger = require 'modules.logger'
 local storage = require 'server.storage.main'
 
 ---@class PlayerData : PlayerEntity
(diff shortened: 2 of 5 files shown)


config/server.lua
```diff
@@ -136,5 +136,8 @@ return {
 
     removeSocietyMoney = function(accountName, payment)
         return exports['Renewed-Banking']:removeAccountMoney(accountName, payment)
-    end
+    end,
+
+    maxJobsPerPlayer = 1,
+    maxGangsPerPlayer = 1,
```
FjamZoo, line 142: Personally I'm more of a fan of using convars for these sort of things?

We haven't really discussed this at length in the developer chat, but would be worth considering

FjamZoo (approved):
Overall good change and is needed.

solareon (approved):
Is an export planned for something like exports.qbx_core:HasJob(playerId) for checking if a player has that job even if they are not actively onduty? I could see this being useful for cases where someone wants to have a secondary job but not be required to switch to it to perform the various job functions

Manason:
> Is an export planned for something like exports.qbx_core:HasJob(playerId) for checking if a player has that job even if they are not actively onduty? I could see this being useful for cases where someone wants to have a secondary job but not be required to switch to it to perform the various job functions

We already have the table on PlayerData, so people can get it via doing GetOfflinePlayer(citizenid)

MafewTM (approved):

config/server.lua
Manason: changed to convar

PR #371, opened 2024-02-18
fix(bridge): core not loading

Manason:
Fixes an issue where the core didn't finish loading before other resources loaded. Doing a lib.callback.await doesn't block other resources from loading, but does block core. To get around this, the jobs/gangs tables of bridge are initialized to empty tables and then will be populated asynchronously once the lib callback returns. The event is then triggered to simulate as if these groups were added via export.

### DIFF
--- a/bridge/qb/client/main.lua
+++ b/bridge/qb/client/main.lua
@@ -6,7 +6,11 @@ require 'bridge.qb.client.events'
 local qbCoreCompat = {}
 qbCoreCompat.PlayerData = QBX.PlayerData
 qbCoreCompat.Config = lib.table.merge(require 'config.client', require 'config.shared')
+
 qbCoreCompat.Shared = require 'bridge.qb.shared.main'
+qbCoreCompat.Shared.Jobs = {}
+qbCoreCompat.Shared.Gangs = {}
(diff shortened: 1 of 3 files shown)


solareon (approved):

bridge/qb/client/main.lua
```diff
@@ -6,7 +6,11 @@ require 'bridge.qb.client.events'
 local qbCoreCompat = {}
 qbCoreCompat.PlayerData = QBX.PlayerData
 qbCoreCompat.Config = lib.table.merge(require 'config.client', require 'config.shared')
+
 qbCoreCompat.Shared = require 'bridge.qb.shared.main'
+qbCoreCompat.Shared.Jobs = {}
+qbCoreCompat.Shared.Gangs = {}
```
D4isDAVID, line 12: Can we not keep these definitions in the main shared file? I'm pretty sure the server needs them too

MafewTM (approved):

bridge/qb/client/main.lua
Manason: Forgot to add the server side. Fixed.

Issue #367, opened 2024-02-15
[Enhancement] built in queue priority system


SKITTLE6969:
### The problem

missing couple of options from the standard queue system

### Ideal solution

add below config options to the queue system may be, if that is not in the bucket list at least add config option to enable/disable built in queue system and use a 3rd party one 

https://github.com/Nick78111/ConnectQueue/blob/b170946921a85640786422fd723ea91bbd23f72d/connectqueue/server/sv_queue_config.lua#L6

https://github.com/Nick78111/ConnectQueue/blob/b170946921a85640786422fd723ea91bbd23f72d/connectqueue/server/sv_queue_config.lua#L31

solareon:
Regarding the disabling of the queue. If you set the convar as described [here](https://docs.qbox.re/converting) then any external queue should work.

D4isDAVID:
The grace time option will be added.

Other than that like @solareon said there is already an option to enable/disable the built-in queue system with the `qbx:enablequeue` convar.

Queue priority using points is not planned at all, instead you should be using the sub-queue system.

D4isDAVID:
Just a note that although a fix has been pushed already the docs above were invalid and you should specify `set qbx:enablequeue "false"`, and *not* `"true"`.

SKITTLE6969:
i saw and tried that but below happens @_@
[     script:qbx_core] [qbx_core] [ERROR] @qbx_core/server/events.lua:128: attempt to index a boolean value (upvalue 'queue')

DevX32:
> i saw and tried that but below happens @_@ [ script:qbx_core] [qbx_core] [ERROR] @qbx_core/server/events.lua:128: attempt to index a boolean value (upvalue 'queue')
are u adding one by one in ur server or whole qbx_core bcz i checked i am not having that issue

D4isDAVID:
That is a real issue that happens when turning off the queue. It'll be fixed.

D4isDAVID:
@SKITTLE6969 disabling the queue has been fixed.

The team also had a little discussion about your suggestion so some sort of a points system will be done after all as well.

SKITTLE6969:
> @SKITTLE6969 disabling the queue has been fixed.
> 
> The team also had a little discussion about your suggestion so some sort of a points system will be done after all as well.

Thank you <3 much appreciated

D4isDAVID:
Don't close the issue yet, we still haven't implemented the suggestion.

sw4m:
[...]

PR #372, opened 2024-02-18
fix(server): error when disabling queue

D4isDAVID:
## Description

Fixes the issue described in https://github.com/Qbox-project/qbx_core/issues/367#issuecomment-1946603716.

Bump ox_lib to a version that includes a fix for returning `false` from required modules (overextended/ox_lib#502), and return `false` from `queue.lua` when it's disabled (`events.lua` already detects when it returns a falsy value).

## Checklist

### DIFF
--- a/server/queue.lua
+++ b/server/queue.lua
@@ -1,4 +1,4 @@
-if GetConvar('qbx:enablequeue', 'true') == 'false' then return end
+if GetConvar('qbx:enablequeue', 'true') == 'false' then return false end
 
 -- Disable hardcap because it kicks the player when the server is full
 
--- a/server/main.lua
+++ b/server/main.lua
@@ -1,5 +1,5 @@
 lib.versionCheck('Qbox-project/qbx_core')
-if not lib.checkDependency('ox_lib', '3.10.0', true) then error() return end
+if not lib.checkDependency('ox_lib', '3.16.3', true) then error() return end
 
 ---@type 'strict'|'relaxed'|'inactive'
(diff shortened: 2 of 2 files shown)


solareon (approved):

server/queue.lua
```diff
@@ -1,4 +1,4 @@
-if GetConvar('qbx:enablequeue', 'true') == 'false' then return end
+if GetConvar('qbx:enablequeue', 'true') == 'false' then return false end
```
Manason, line 1: Why is it better to return false?

D4isDAVID: Feels much nicer and more readable to be able to do a simple `if not queue` check than checking if the table is empty or if the required module returns some other arbitrary value.

MafewTM (approved):

server/queue.lua
Manason: what happens if you return nil? if not queue doesn't work?

D4isDAVID: Returning nil/nothing will result in the `require` function to return `true`, which I didn't know when writing, and `events.lua` conducts `if not queue` checks because I wrongly expected `require` to return nil.

Manason: What's the ox_lib change then?

D4isDAVID: Previous ox_lib turned any falsy return values from modules to `true`, when only nil values were supposed to become `true`. As mentioned in the description of this PR I bumped to a version that includes the fix for that, so that returning `false` works correctly.

PR #374, opened 2024-02-19
fix(server/player): use job/gang data as source of truth instead of playerdata

Manason:
The only thing the stored playerdata.job and .gang will be used for now is the name, grade, and last onduty value. Other fields of job/gang will be derived from the job/gang data. This fixes the potential for stale cached data on the player object. This doesn't change what data is written to the database.

### DIFF
--- a/server/player.lua
+++ b/server/player.lua
@@ -390,32 +390,40 @@ function CheckPlayerData(source, playerData)
         InstalledApps = {},
     }
     local jobs, gangs = storage.fetchPlayerGroups(playerData.citizenid)
-    -- Job
-    if playerData.job and playerData.job.name and not GetJob(playerData.job.name) then playerData.job = nil end
-    playerData.job = playerData.job or {}
-    playerData.job.name = playerData.job.name or 'unemployed'
-    playerData.job.label = playerData.job.label or 'Civilian'
-    playerData.job.payment = playerData.job.payment or 10
-    playerData.job.type = playerData.job.type or 'none'
-    if QBX.Shared.ForceJobDefaultDutyAtLogin or playerData.job.onduty == nil then
-        playerData.job.onduty = GetJob(playerData.job.name).defaultDuty
+
+    local job = GetJob(playerData.job?.name) or GetJob('unemployed')
+    if not job then
+        error("unemployed job not found. Is it in your config and/or database?")
     end
-    playerData.job.isboss = playerData.job.isboss or false
-    playerData.job.grade = playerData.job.grade or {}
-    playerData.job.grade.name = playerData.job.grade.name or 'Freelancer'
-    playerData.job.grade.level = playerData.job.grade.level or 0
+    local jobGrade = GetJob(playerData.job?.name) and playerData.job.grade.level or 0
+    playerData.job = {
+        name = playerData.job.name or 'unemployed',
+        label = job.label,
+        payment = job.grades[jobGrade].payment or 0,
+        type = job.type,
+        onduty = QBX.Shared.ForceJobDefaultDutyAtLogin and job.defaultDuty or playerData.job.onduty,
+        isboss = job.grades[jobGrade].isboss or false,
+        grade = {
+            name = job.grades[jobGrade].name,
+            level = jobGrade,
+        }
+    }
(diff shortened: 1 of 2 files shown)


solareon (approved):
Clean

server/player.lua
```diff
@@ -390,32 +390,40 @@ function CheckPlayerData(source, playerData)
...
-    playerData.job.isboss = playerData.job.isboss or false
-    playerData.job.grade = playerData.job.grade or {}
-    playerData.job.grade.name = playerData.job.grade.name or 'Freelancer'
-    playerData.job.grade.level = playerData.job.grade.level or 0
+
+    local job = GetJob(playerData.job?.name) or GetJob('unemployed')
+    if not job then
+        error("unemployed job not found. Is it in your config and/or database?")
+    end
+    local jobGrade = GetJob(playerData.job?.name) and playerData.job.grade.level or 0
```
BerkieBb, line 398: GetJob is redundant here and job grade should be grabbed from the `job` variable

server/player.lua
```diff
@@ -390,32 +390,40 @@ function CheckPlayerData(source, playerData)
...
-    playerData.gang.label = playerData.gang.label or 'No Gang Affiliation'
-    playerData.gang.isboss = playerData.gang.isboss or false
-    playerData.gang.grade = playerData.gang.grade or {}
-    playerData.gang.grade.name = playerData.gang.grade.name or 'none'
-    playerData.gang.grade.level = playerData.gang.grade.level or 0
+    local gang = GetGang(playerData.gang?.name) or GetGang('none')
+    if not gang then
+        error("none gang not found. Is it in your config and/or database?")
+    end
+    local gangGrade = GetGang(playerData.gang?.name) and playerData.gang.grade.level or 0
```
BerkieBb, line 416: Same here as job stuff

server/storage/main.lua
```diff
@@ -24,7 +24,7 @@ local groups = require 'server.storage.groups'
 ---@field fetchIsUnique fun(type: UniqueIdType, value: string|number): boolean
 ---@field addPlayerToJob fun(citizenid: string, group: string, grade: integer)
 ---@field addPlayerToGang fun(citizenid: string, group: string, grade: integer)
----@field fetchPlayerGroups fun(citizenid: string): table<string, Job>, table<string, Gang>
+---@field fetchPlayerGroups fun(citizenid: string): table<string, integer>, table<string, integer> jobs, gangs
```
BerkieBb, line 27: Why this change and why the jobs, gangs at the end?

server/player.lua
solareon:
```suggestion
    local jobGrade = job.grade[playerData.job?.grade.level].level or 0
```
I believe this is what Berkie means. This should cover cases where a grade is deleted and the player has a grade that no longer exists

server/player.lua
solareon:
```suggestion
    local gangGrade = gang.grade[playerData.gang?.grade.level].level or 0
```

server/storage/main.lua
Manason: Trying to communicate that the first table are jobs, and the second table are gangs. I guess the non lazy way to do it is with aliases.

This is a fix to make the documentation match the actual return value of the function

server/player.lua
Manason: the job variable is the actual job itself, not the player's job. Therefore it doesn't have player specific grade info, but rather all the grades of the job instead.

Manason: > I believe this is what Berkie means. This should cover cases where a grade is deleted and the player has a grade that no longer exists

The reason why this won't work (aside from some off var names) is that job is either the saved job, or unemployed. If it's unemployed we want to disregard the saved player grade and go with 0.

solareon: > > I believe this is what Berkie means. This should cover cases where a grade is deleted and the player has a grade that no longer exists
> 
> The reason why this won't work (aside from some off var names) is that job is either the saved job, or unemployed. If it's unemployed we want to disregard the saved player grade and go with 0.

My bad. I thought we were assembling the client side. disregard my comment

FjamZoo (approved):

Issue #384, opened 2024-02-26
Incorrect string value for column name (DB encoding)


nitrou5:
### Summary

DB failure to save players with unicode emojis in their name

### Reproduction

Found this reported on discord, was able to reproduce it and confirm:
1. Include emoji in your name in FiveM settings
2. Join Server for first time to trigger insert player row in table
3. See Incorrect string value for column name players.name
![image](https://github.com/Qbox-project/qbx_core/assets/157820969/704e5c3c-f108-4731-9bb0-73f9f5322986)
![image](https://github.com/Qbox-project/qbx_core/assets/157820969/448922d3-2db8-410f-8cab-619a721c2884)

### Expected behavior

No server errors, player need to persist in DB table

### Actual behavior

player with emoji in their name do not persist in DB table

### Current Version

v1.6.1

### Custom Resources

n/a

PR #385, opened 2024-02-26
default qbx_core SQL tables to utf8mb4 charset

nitrou5:
## Description

default core SQL tables to utf8mb4 charset to prevent encoding issues in qbx_core.sql
Solves #384 

## Checklist

### DIFF
--- a/qbx_core.sql
+++ b/qbx_core.sql
@@ -16,7 +16,7 @@ CREATE TABLE IF NOT EXISTS `players` (
   KEY `id` (`id`),
   KEY `last_updated` (`last_updated`),
   KEY `license` (`license`)
-) ENGINE=InnoDB AUTO_INCREMENT=1;
+) ENGINE=InnoDB AUTO_INCREMENT=1 DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;
 
 CREATE TABLE IF NOT EXISTS `bans` (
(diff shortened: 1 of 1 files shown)


Manason (changes requested):
What about player_groups?

Manason:
> I am not sure if this is better than sanitizing/purifying the input from client its up to you the final decision

It seems correct to me that we should support utf8

Manason (approved):

qbx_core.sql
```diff
@@ -11,12 +11,13 @@ CREATE TABLE IF NOT EXISTS `players` (
   `position` text NOT NULL,
   `metadata` text NOT NULL,
   `inventory` longtext DEFAULT NULL,
+  `phone_number` VARCHAR(20) DEFAULT NULL,
```
TonybynMp4, line 14: Do we want to add the phone_number table in core?

Manason: Oh, I didn't even see this. This doesn't seem related to the PR description

Manason (changes requested):

TonybynMp4 (approved):

Manason (approved):

qbx_core.sql
nitrou5: My bad was using the recipe SQL as a reference, its also out of scope for this PR
https://github.com/Qbox-project/txAdminRecipe/blob/main/qbox.sql#L133 thanks for catching that

PR #388, opened 2024-02-28
feat: add vehicle shops back

TonybynMp4:
## Description

Bring the shop field back, added the shops from the vehicleshop config with a script, that's why for some vehicles the order might be corrected (also the trailing commas are cleaned up)

Some team members might be against this idk where we left this off..
## Checklist


Manason:
Convince the team before opening PRs on this.

PR #389, opened 2024-02-28
chore(server): deprecate permissions in favor of cfg ACE

Manason:
The abstraction built into qb on top of ACE permissions is confusing without adding value. Deprecated with the recommendation that developers use native ACE permissions + ox_lib to manage permissions.

### DIFF
--- a/server/commands.lua
+++ b/server/commands.lua
@@ -98,7 +98,7 @@ lib.addCommand('openserver', {
         Notify(source, locale('error.server_already_open'), 'error')
         return
     end
-    if HasPermission(source, 'admin') then
+    if IsPlayerAceAllowed(source, 'admin') then
         config.server.closed = false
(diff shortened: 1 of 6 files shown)


server/commands.lua
solareon, line 1: Should we also mark the addCommands as deprecated since they depend on the deprecated functions?

MafewTM (approved):

FjamZoo (approved):

server/commands.lua
Manason: Maybe. Or maybe we should keep them and change their behavior to be wrappers for ox_lib? Wdyt?

solareon: I presume that the ox_lib variety is also runtime only and lost upon restart? If we keep them though using the ox_lib versions would be the most ideal.

PR #394, opened 2024-03-01
feat: lastLoggedOut in PlayerData

TonybynMp4:
## Description
Adds a new db column 'lastLoggedOut' that is updated on logout & playerDropped.
This new value can be accessed through the PlayerData & used to know when the player last logged out.

## Checklist

### DIFF
--- a/qbx_core.sql
+++ b/qbx_core.sql
@@ -12,6 +12,7 @@ CREATE TABLE IF NOT EXISTS `players` (
   `metadata` text NOT NULL,
   `inventory` longtext DEFAULT NULL,
   `phone_number` VARCHAR(20) DEFAULT NULL,
+  `lastLoggedOut` timestamp NOT NULL DEFAULT current_timestamp(),
   `last_updated` timestamp NOT NULL DEFAULT current_timestamp() ON UPDATE current_timestamp(),
   PRIMARY KEY (`citizenid`),
   KEY `id` (`id`),
--- a/qbx_core.sql
+++ b/qbx_core.sql
@@ -12,7 +12,7 @@ CREATE TABLE IF NOT EXISTS `players` (
   `metadata` text NOT NULL,
   `inventory` longtext DEFAULT NULL,
   `phone_number` VARCHAR(20) DEFAULT NULL,
-  `lastLoggedOut` timestamp NOT NULL DEFAULT current_timestamp(),
+  `lastLoggedOut` timestamp DEFAULT NULL,
   `last_updated` timestamp NOT NULL DEFAULT current_timestamp() ON UPDATE current_timestamp(),
   PRIMARY KEY (`citizenid`),
   KEY `id` (`id`),
--- a/server/events.lua
+++ b/server/events.lua
@@ -40,6 +40,7 @@ AddEventHandler('playerDropped', function(reason)
     if not QBX.Players[src] then return end
     GlobalState.PlayerCount -= 1
     local player = QBX.Players[src]
+    player.PlayerData.lastLoggedOut = os.date('%Y-%m-%d %H:%M:%S')
     logger.log({
         source = 'qbx_core',
         webhook = loggingConfig.webhook['joinleave'],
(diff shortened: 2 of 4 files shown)


MafewTM (approved):

server/events.lua
```diff
@@ -40,6 +40,7 @@ AddEventHandler('playerDropped', function(reason)
     if not QBX.Players[src] then return end
     GlobalState.PlayerCount -= 1
     local player = QBX.Players[src]
+    player.PlayerData.lastLoggedOut = os.time() * 1000
```
Manason, line 43: I don't think you should need to multiply here. I assume you are trying to add the milliseconds, but is that needed for os.date?

server/player.lua
```diff
@@ -425,6 +425,7 @@ function Logout(source)
 
     local player = GetPlayer(source)
     if not player then return end
+    player.PlayerData.lastLoggedOut = os.time() * 1000
```
Manason, line 428: ditto

server/storage/players.lua
```diff
@@ -198,6 +199,7 @@ local function fetchAllPlayerEntities(license2, license)
         chars[i].gang = result[i].gang and json.decode(result[i].gang)
         chars[i].position = convertPosition(result[i].position)
         chars[i].metadata = json.decode(result[i].metadata)
+        chars[i].lastLoggedOut = result[i].lastLoggedOut and result[i].lastLoggedOut/1000
```
Manason, line 202: math.floor the division?

server/storage/players.lua
```diff
@@ -219,7 +221,8 @@ local function fetchPlayerEntity(citizenId)
         job = player.job and json.decode(player.job),
         gang = player.gang and json.decode(player.gang),
         position = convertPosition(player.position),
-        metadata = json.decode(player.metadata)
+        metadata = json.decode(player.metadata),
+        lastLoggedOut = player.lastLoggedOut/1000
```
Manason, line 225: ditto

Manason (changes requested):

Manason:
Did you test this?

qbx_core.sql
```diff
@@ -12,6 +12,7 @@ CREATE TABLE IF NOT EXISTS `players` (
   `metadata` text NOT NULL,
   `inventory` longtext DEFAULT NULL,
   `phone_number` VARCHAR(20) DEFAULT NULL,
+  `lastLoggedOut` timestamp NOT NULL DEFAULT current_timestamp(),
```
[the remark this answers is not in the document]
TonybynMp4: I mean duh

server/storage/players.lua
TonybynMp4: no need to floor a whole number

BerkieBb: It isn't a whole number always when you divide

TonybynMp4: Ok nvm in this case we shouldn't even divide since MYSQL.query returns Milliseconds and not microseconds for some reason

server/storage/players.lua
```diff
@@ -198,6 +201,7 @@ local function fetchAllPlayerEntities(license2, license)
         chars[i].gang = result[i].gang and json.decode(result[i].gang)
         chars[i].position = convertPosition(result[i].position)
         chars[i].metadata = json.decode(result[i].metadata)
+        chars[i].lastLoggedOut = result[i].lastLoggedOut and result[i].lastLoggedOut
```
Manason, line 204:
```suggestion
        chars[i].lastLoggedOut = result[i].lastLoggedOut
```

server/storage/players.lua
Manason: This is different than fetching all player entities? Also you need to math.floor the division to guarantee that it won't be a decimal number

Manason (changes requested):

server/storage/players.lua
TonybynMp4: My bad for the difference,
Still no need to floor it tho but ok.

server/storage/players.lua
```diff
@@ -99,6 +100,7 @@ end
 ---@field position vector4
 ---@field metadata PlayerMetadata
 ---@field cid integer
+---@field LastLoggedOut integer
```
Manason, line 103:
```suggestion
---@field lastLoggedOut integer
```
Should be camelCase everywhere as well

PR #395, opened 2024-03-01
feat: use statebags for hunger/thirst/stress

TonybynMp4:
## Description
Statebags are cool :+1:
It keeps compatibility with the old way of setting those values, so we can start using statebags without anything being broken until then!

(please test to make sure i didn't fuck it up)

## Checklist

### DIFF
--- a/bridge/qb/server/events.lua
+++ b/bridge/qb/server/events.lua
@@ -39,3 +39,12 @@ RegisterServerEvent('baseevents:leftVehicle', function(veh, seat, modelName, net
     }
     TriggerClientEvent('QBCore:Client:VehicleInfo', src, data)
 end)
+
+-- Player
+RegisterServerEvent('qbx_core:server:onSetMetaData', function (meta, _, val, source)
+    if meta == 'thirst' or meta == 'hunger' then return end
+    local playerState = Player(source).state
+    if val == playerState[meta] then return end
+    if not playerState.isLoggedIn then return end
+    playerState:set(meta, val, true)
+end)
--- a/bridge/qb/server/events.lua
+++ b/bridge/qb/server/events.lua
@@ -39,12 +39,3 @@ RegisterServerEvent('baseevents:leftVehicle', function(veh, seat, modelName, net
     }
     TriggerClientEvent('QBCore:Client:VehicleInfo', src, data)
 end)
-
--- Player
-RegisterServerEvent('qbx_core:server:onSetMetaData', function (meta, _, val, source)
-    if meta == 'thirst' or meta == 'hunger' then return end
-    local playerState = Player(source).state
-    if val == playerState[meta] then return end
-    if not playerState.isLoggedIn then return end
-    playerState:set(meta, val, true)
-end)
--- a/server/events.lua
+++ b/server/events.lua
@@ -204,7 +204,7 @@ RegisterNetEvent('QBCore:ToggleDuty', function()
 end)
 
 RegisterServerEvent('qbx_core:server:onSetMetaData', function (meta, _, val, source)
-    if meta ~= 'thirst' or meta ~= 'hunger' then return end
+    if meta ~= 'thirst' and meta ~= 'hunger' then return end
     local playerState = Player(source).state
     if val == playerState[meta] then return end
     if not playerState.isLoggedIn then return end
--- a/server/events.lua
+++ b/server/events.lua
@@ -203,7 +203,7 @@ RegisterNetEvent('QBCore:ToggleDuty', function()
 end)
 
 RegisterServerEvent('qbx_core:server:onSetMetaData', function (meta, _, val, source)
-    if meta == 'thirst' or meta == 'hunger' then return end
+    if meta ~= 'thirst' or meta ~= 'hunger' then return end
     local playerState = Player(source).state
     if val == playerState[meta] then return end
     if not playerState.isLoggedIn then return end
--- a/server/events.lua
+++ b/server/events.lua
@@ -202,11 +202,3 @@ RegisterNetEvent('QBCore:ToggleDuty', function()
     end
     TriggerClientEvent('QBCore:Client:SetDuty', src, player.PlayerData.job.onduty)
 end)
-
-RegisterServerEvent('qbx_core:server:onSetMetaData', function (meta, _, val, source)
-    if meta ~= 'thirst' and meta ~= 'hunger' then return end
-    local playerState = Player(source).state
-    if val == playerState[meta] then return end
-    if not playerState.isLoggedIn then return end
-    playerState:set(meta, val, true)
-end)
--- a/server/events.lua
+++ b/server/events.lua
@@ -201,3 +201,11 @@ RegisterNetEvent('QBCore:ToggleDuty', function()
     end
     TriggerClientEvent('QBCore:Client:SetDuty', src, player.PlayerData.job.onduty)
 end)
+
+RegisterServerEvent('qbx_core:server:onSetMetaData', function (meta, _, val, source)
+    if meta == 'thirst' or meta == 'hunger' then return end
+    local playerState = Player(source).state
+    if val == playerState[meta] then return end
+    if not playerState.isLoggedIn then return end
+    playerState:set(meta, val, true)
+end)
--- a/server/loops.lua
+++ b/server/loops.lua
@@ -6,8 +6,9 @@ local function removeHungerAndThirst(src, player)
     local newHunger = playerState.hunger - config.player.hungerRate
     local newThirst = playerState.thirst - config.player.thirstRate
 
-    playerState:set('hunger', math.max(0, newHunger), true)
-    playerState:set('thirst', math.max(0, newThirst), true)
+    player.Functions.SetMetaData('thirst', math.max(0, newHunger))
+    player.Functions.SetMetaData('hunger', math.max(0, newThirst))
+
     TriggerClientEvent('hud:client:UpdateNeeds', src, newHunger, newThirst)
     player.Functions.Save()
 end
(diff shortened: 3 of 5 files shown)


bridge/qb/server/events.lua
```diff
@@ -38,4 +38,13 @@ RegisterServerEvent('baseevents:leftVehicle', function(veh, seat, modelName, net
...
     TriggerClientEvent('QBCore:Client:VehicleInfo', src, data)
+end)
+
+-- Player
+RegisterServerEvent('qbx_core:server:onSetMetaData', function (meta, _, val, source)
+    if meta == 'thirst' or meta == 'hunger' then return end
+    local playerState = Player(source).state
+    if val == playerState[meta] then return end
+    if not playerState.isLoggedIn then return end
+    playerState:set(meta, val, true)
```
Manason, line 49: Why is this in bridge. Shouldn't this be within the main core files?

server/loops.lua
```diff
@@ -1,16 +1,18 @@
...
+    local newHunger = playerState.hunger - config.player.hungerRate
+    local newThirst = playerState.thirst - config.player.thirstRate
     if newHunger <= 0 then
         newHunger = 0
     end
     if newThirst <= 0 then
         newThirst = 0
     end
-    player.Functions.SetMetaData('thirst', newThirst)
-    player.Functions.SetMetaData('hunger', newHunger)
```
Manason, line 13: Don't we need to continue to keep the metadata updated?

server/player.lua
```diff
@@ -420,11 +420,16 @@ end
 ---On player logout
 ---@param source Source
 function Logout(source)
+    local player = GetPlayer(source)
+    if not player then return end
+    local playerState = Player(source)?.state
```
Manason, line 425: If the player drops, won't the state bag be destroyed?

Manason (changes requested):

server/player.lua
TonybynMp4: It's not destroyed, at least in the playerDropped event we still have access to the statebag.

bridge/qb/server/events.lua
TonybynMp4: Because it's Backwards compatibility with Scripts that still use the setMetadata way? (ours still do but it's temporary)

server/loops.lua
TonybynMp4: It's saved anyway,
if you log out / drop it's saved,
when it's updated via the loop it's saved.

So i think it's safe to not use SetMetadata everytime.

bridge/qb/server/events.lua
Manason: The bridge is for backwards compatibility with qb. This is backwards compatibility with previous versions of Qbox.

TonybynMp4: > The bridge is for backwards compatibility with qb. This is backwards compatibility with previous versions of Qbox.

It's backwards compat with qb, and qbox.

server/loops.lua
MafewTM: I would test this so we know

bridge/qb/server/events.lua
Manason: Shouldn't someone be able to turn off the bridge using config if they aren't getting the core object? The idea of the bridge is to provide backwards compatibility to qb, not previous versions of Qbox.

TonybynMp4: Hmm ok i see, i'll move it to regular server/events

Faprwnt:
I tried to test this, there was an error on line 10 about compare nil with number and in the console it was arithemetic on nil value server/loops.lua:6

TonybynMp4:
> I tried to test this, there was an error on line 10 about compare nil with number and in the console it was arithemetic on nil value server/loops.lua:6

Try again now, errors makes sense if the values aren't set to begin with 😆

Fraprwnt:
> > I tried to test this, there was an error on line 10 about compare nil with number and in the console it was arithemetic on nil value server/loops.lua:6
> 
> Try again now, errors makes sense if the values aren't set to begin with 😆

Yes, this works well, there are no errors, I found several problems, when I eat it does increase in the bar, but when I continue eating it gets stuck and doesn't add any more, and when I drink it's the same, what's there actually decreases.

TonybynMp4:
> when I eat it does increase in the bar, but when I continue eating it gets stuck and doesn't add any more, and when I drink it's the same, what's there actually decreases.

That's weird, i don't really see why that would happen, tbh 🤔

MafewTM:
I'm experiencing the same issue too. When I use this with your hud pr, the first time I eat it updates but after that, it doesnt update anymore from what I can tell.

TonybynMp4:
> Yes, this works well, there are no errors, I found several problems, when I eat it does increase in the bar, but when I continue eating it gets stuck and doesn't add any more, and when I drink it's the same, what's there actually decreases.

Can you please update core smallresources & hud and try again now? :)
(you too @MafewTM)

Fraprwnt:
> > Yes, this works well, there are no errors, I found several problems, when I eat it does increase in the bar, but when I continue eating it gets stuck and doesn't add any more, and when I drink it's the same, what's there actually decreases.
> 
> @Fraprwnt can you please update core smallresources & hud and try again now? :) (you too @MafewTM)

to eat using items from ox_inventory remains the same. It only increases after the first meal. The second and third don't change anything, but to eat using items from the smallresources config, for the first and second meal the same amount of food and drink increases.

server/loops.lua
```diff
@@ -1,16 +1,18 @@
...
+    playerState:set('hunger', newHunger, true)
+    playerState:set('thirst', newThirst, true)
```
FjamZoo, line 15: we should use math.min here instead of wierd if statements to check if the value is less than 0

```lua
playerState:set('hunger', math.min(0, newHunger), true)
playerState:set('thirst', math.min(0, newThirst), true)
```

server/player.lua
```diff
@@ -420,18 +425,26 @@ end
...
     player.PlayerData.lastLoggedOut = os.time()
     player.Functions.Save()
 
     Wait(200)
     QBX.Players[source] = nil
     GlobalState.PlayerCount -= 1
     TriggerClientEvent('qbx_core:client:playerLoggedOut', source)
+    playerState:set("stress", 100, true)
+    playerState:set("thirst", 100, true)
+    playerState:set("hunger", 100, true)
```
FjamZoo, line 447: Single quotes please.

server/player.lua
```diff
@@ -335,6 +336,10 @@ function CheckPlayerData(source, playerData)
     playerData.metadata.hunger = playerData.metadata.hunger or 100
     playerData.metadata.thirst = playerData.metadata.thirst or 100
     playerData.metadata.stress = playerData.metadata.stress or 0
+    playerState:set("hunger", playerData.metadata.hunger, true)
+    playerState:set("thirst", playerData.metadata.thirst, true)
+    playerState:set("stress", playerData.metadata.stress, true)
```
FjamZoo, line 341: SINGLE QUOTES

server/events.lua
```diff
@@ -202,3 +202,11 @@ RegisterNetEvent('QBCore:ToggleDuty', function()
     end
     TriggerClientEvent('QBCore:Client:SetDuty', src, player.PlayerData.job.onduty)
 end)
+
+RegisterServerEvent('qbx_core:server:onSetMetaData', function (meta, _, val, source)
```
FjamZoo, line 206: Why do we use this event instead of using the internal update metadata function?

FjamZoo (changes requested):

server/events.lua
TonybynMp4: Hum true, idk if what i changed is what you meant for sure tho

server/player.lua
```diff
@@ -542,6 +555,7 @@ function CreatePlayer(playerData, Offline)
         if not meta or type(meta) ~= 'string' then return end
         if meta == 'hunger' or meta == 'thirst' then
             val = val > 100 and 100 or val
+            Player(self.PlayerData.source).state:set(meta, val, true)
```
lwhp, line 558: I know this was not really part of the original pr but this 
```lua
val = val > 100 and 100 or val
```
Is extremely messy and we have (thanks to ox_lib) a better math library therefor please use lib.math.clamp here.

```lua
Player(self.PlayerData.source).state:set(meta, math.clamp(val, 0, 100), true)
```

FjamZoo (approved):

server/loops.lua
Manason: We need to continue to trigger the metadata set event when metadata is changed to avoid breaking our API contract https://docs.qbox.re/resources/qbx_core/events/server#qbx_coreserveronsetmetadata

Manason (changes requested):
metadata must always be modified using the SetMetadata function of player. Otherwise certain events won't fire, which are relied upon for updating QBX.PlayerData for example, and likely used in other scripts.

Manason (approved):
Fix listing errors and make sure this is thoroughly tested please

mafewtm (approved):
As long as it's been thoroughly tested.

server/player.lua
```diff
@@ -554,8 +567,8 @@ function CreatePlayer(playerData, Offline)
     ---@param val any
     function self.Functions.SetMetaData(meta, val)
         if not meta or type(meta) ~= 'string' then return end
-        if meta == 'hunger' or meta == 'thirst' then
-            val = val > 100 and 100 or val
+        if meta == 'hunger' or meta == 'thirst' or meta == 'stress' then
+            Player(self.PlayerData.source).state:set(meta, math.clamp(val, 0, 100), true)
```
mafewtm, line 571: Pretty sure you gotta use `math = lib.math` with this right?

TonybynMp4: Works fine for me 🤔 🤷🏻

PR #404, opened 2024-03-09
Reformat the code

bicatcho11:
## Description

This formatting makes the code easier to read and understand, especially for someone who might be reviewing or modifying it later.

## Checklist

### DIFF
--- a/shared/gangs.lua
+++ b/shared/gangs.lua
@@ -1,125 +1,69 @@
 ---@type table<string, Gang>
 return {
-	['none'] = {
-		label = 'No Gang',
-		grades = {
-            [0] = {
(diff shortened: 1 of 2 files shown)


D4isDAVID:
I believe this does the opposite of what is intended. With these changes the files do become smaller and more concise, but that doesn't necessarily mean they're more readable.

Having *all* fields of a table defined on the same line is less readable, less modifiable, and can show up weird in diffs when committing changes. It gets especially annoying when lines defining tables get too long. [`items.lua`](https://github.com/Qbox-project/qbx_core/blob/2b1af1aef2a1540335d821c57fd309a789829feb/shared/items.lua) is a great example for this.

PR #409, opened 2024-03-11
feat: setjob replace convar

Manason:
- Adds two new convars qbx:setjob_replace and qbx:setgang_replace
- These determine whether the player function SetJob & SetGang should replace your existing primary job/gang or not. They default to true, but some users may want to turn the replacement off.


Randolio:
Tested this on an up to date build. Works perfectly for what was discussed between us in qbox discord. Server owners just need to make sure they set their convars correctly, including increasing the max jobs limit with setJobReplaces being false.

solareon (approved):

FjamZoo (approved):

PR #412, opened 2024-03-12
refactor: vehicle props as statebag

### LINKED ISSUE #316: Vehicle spawning irregularities
### Summary

Near daily reports of vehicles requested from garage spawning with incorrect plate and missing mods

### Reproduction

I have spent over 8 hours trying to reproduce and cannot reproduce locally or from remote server. Players are requesting a personal vehicle from their garage and this will happen sometimes, usually once per day at approx. 15 avg players. Has happened from v1.2+. Cannot verify if happened prior to 1.2.

### Expected behavior

The plate to match the vehicle 100% of the time, keys given and mods to always apply to the vehicle.

### Actual behavior

In small cases (<5% observed) the plate will not set. 

1. Using qbx_garages or rahe-boosting (I implemented server side vehicle spawn from QBX), a vehicle is created
2. The vehicle's plate will not match the requested vehicle's plate
3. User cannot drive vehicle (no keys), and their cosmetics and performance mods are not applied

### Additional context

My server is a low pop beta with 10-20 players online when the issue generally occurs. 

I created a thread in the Discord with my debugging information as I went through the process:
https://discord.com/channels/1012753553418354748/1195501924041117727

There were other issues discovered that were originally believed to be related (same issue happening with job spawns), but I believe those are unrelated now. **At least for qbx_ambulancejob,** the missing keys/wrong plate is believed to be due to the plate being set on the requestor's client without regard to who the entity owner is. The warning you'd expect is masked because the properties are not set by ox_lib as they are in the core.

I understand without reproduction steps that this is a long shot but it is a very troublesome issue. I will appreciate any suggestions or research into this.

### Last Updated

1-12-2023

### Custom Resources

Yes but nothing that will interfere with standard vehicle spawning

### Resource Rename

No

solareon:
## Description

Vehicle props are sadly a client side thing. This change moves the logic from a callback to a statebag which runs a loop on any client within scope to set the props on a vehicle. Extensive testing has shown this is the best method of ensuring that if you ask for a blue car with a wing that you get a blue car with a wing no matter who is zipping by at 400mph or if your client/server is a potato.

Fixes #316

Credit @FjamZoo for the ox_core version of this

## Checklist

### DIFF
--- a/client/events.lua
+++ b/client/events.lua
@@ -196,17 +196,23 @@ RegisterNetEvent('QBCore:Client:OnSharedUpdateMultiple', function(tableName, val
 end)
 
 -- Set vehicle props
----@param entity number
+---@param vehicle number
 ---@param props table<any, any>
-qbx.entityStateHandler('setVehicleProperties', function(entity, _, props)
+qbx.entityStateHandler('setVehicleProperties', function(vehicle, _, props)
     if not props then return end
 
     SetTimeout(0, function()
-        local state = Entity(entity).state
+        local state = Entity(vehicle).state
 
         while state.setVehicleProperties do
-            if NetworkGetEntityOwner(entity) == cache.playerId then
-                if lib.setVehicleProperties(entity, props) then
+            if NetworkGetEntityOwner(vehicle) == cache.playerId then
+                for i = -1, 0 do
+                    local ped = GetPedInVehicleSeat(vehicle, i)
+                    if ped ~= cache.ped and ped > 0 then
+                        DeleteEntity(ped)
+                    end
+                end
+                if lib.setVehicleProperties(vehicle, props) then
                     state:set('setVehicleProperties', nil, true)
                 end
             end
--- a/client/events.lua
+++ b/client/events.lua
@@ -204,20 +204,47 @@ qbx.entityStateHandler('setVehicleProperties', function(vehicle, _, props)
     SetTimeout(0, function()
         local state = Entity(vehicle).state
 
+        local timeOut = GetGameTimer() + 10000
+
         while state.setVehicleProperties do
             if NetworkGetEntityOwner(vehicle) == cache.playerId then
-                for i = -1, 0 do
-                    local ped = GetPedInVehicleSeat(vehicle, i)
-                    if ped ~= cache.ped and ped > 0 then
-                        DeleteEntity(ped)
-                    end
-                end
                 if lib.setVehicleProperties(vehicle, props) then
                     state:set('setVehicleProperties', nil, true)
                 end
             end
+            if GetGameTimer() > timeOut then
+                break
+            end
 
-            Wait(0)
+            Wait(50)
         end
+    end)
+end)
+
+-- Clear vehicle peds
+---@param vehicle number
+---@param init boolean
+qbx.entityStateHandler('initVehicle', function(vehicle, _, init)
+    if not init then return end
+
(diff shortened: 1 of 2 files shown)


client/events.lua
```diff
@@ -129,20 +129,6 @@ RegisterNetEvent('QBCore:Command:GoToMarker', function()
...
 
 -- Vehicle Commands
-lib.callback.register('qbx_core:client:vehicleSpawned', function(netId, props)
-    local veh = NetworkGetEntityFromNetworkId(netId)
-
-    for i = -1, 0 do
-        local ped = GetPedInVehicleSeat(veh, i)
-        if ped ~= cache.ped and ped > 0 and NetworkGetEntityOwner(ped) == cache.playerId then
-            DeleteEntity(ped)
-        end
```
Manason, line 139: Vehicles sometimes spawn with NPCs which this took care of deleting. Can we move this logic over as well?

Manason (approved):

client/events.lua
solareon: done

solareon:
> Does this still require modification of ox_lib (uncommenting a few lines as mentioned in #316)?

@jnccloud No this moves the modification into the core while still using the ox_lib functionally to do the actual prop settings

modules/lib.lua
```diff
@@ -260,13 +260,15 @@ if isServer then
             SetPedIntoVehicle(ped, veh, -1)
         end
 
+        ---@diagnostic disable-next-line: unused-local
         local owner = lib.waitFor(function()
             local owner = NetworkGetEntityOwner(veh)
             if owner ~= -1 then return owner end
         end, 5000)
```
FjamZoo, line 267: Lets get rid of this terrible fucking code thanks.

client/events.lua
```diff
@@ -208,3 +194,30 @@ RegisterNetEvent('QBCore:Client:OnSharedUpdateMultiple', function(tableName, val
...
+        local state = Entity(vehicle).state
+
+        while state.setVehicleProperties do
+            if NetworkGetEntityOwner(vehicle) == cache.playerId then
+                for i = -1, 0 do
+                    local ped = GetPedInVehicleSeat(vehicle, i)
+                    if ped ~= cache.ped and ped > 0 then
+                        DeleteEntity(ped)
+                    end
+                end
```
FjamZoo, line 214: This is very flawed, just because you are the entity owner of the car dosnt mean u are the entity owner of the ped and therefor cant delete it.

Also shouldnt be used in a loop

modules/lib.lua
```diff
@@ -260,14 +261,11 @@ if isServer then
             SetPedIntoVehicle(ped, veh, -1)
         end
 
-        local owner = lib.waitFor(function()
-            local owner = NetworkGetEntityOwner(veh)
-            if owner ~= -1 then return owner end
-        end, 5000)
-
+        Entity(veh).state:set('setVehicleProperties', props, true)
+        Entity(veh).state:set('initVehicle', true, true)
```
FjamZoo, line 265:
```suggestion
        Entity(veh).state:set('initVehicle', true, true)
        Entity(veh).state:set('setVehicleProperties', props, true)
```

We need to intiate the vehicle before trying to set the properties

FjamZoo (changes requested):

FjamZoo (approved):

PR #417, opened 2024-03-24
refactor: set props before ped warp

solareon:
## Description

When warping a ped into a vehicle the vehicle props should already be set otherwise they may fail to set because of entity ownership changes. This moves the prop setting above warping the ped and yields until props are set.

Tested this with two folks with over 100ms latency to the server standing next to each other and props didn't fail to set at all. The wait is for up to 5000ms due to possible issues with streamed vehicles not loading for clients on slow links.

## Checklist

### DIFF
--- a/modules/lib.lua
+++ b/modules/lib.lua
@@ -257,11 +257,11 @@ if isServer then
         while not DoesEntityExist(veh) do Wait(0) end
         while GetVehicleNumberPlateText(veh) == '' do Wait(0) end
 
-        Entity(veh).state:set('initVehicle', true, true)
-        Entity(veh).state:set('setVehicleProperties', props, true)
+        local state = Entity(veh).state
+        state:set('initVehicle', true, true)
+        state:set('setVehicleProperties', props, true)
 
         lib.waitFor(function()
-            local state = Entity(veh).state
             if state.setVehicleProperties then return false end
             return true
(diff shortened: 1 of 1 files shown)


modules/lib.lua
```diff
@@ -257,12 +257,18 @@ if isServer then
         while not DoesEntityExist(veh) do Wait(0) end
         while GetVehicleNumberPlateText(veh) == '' do Wait(0) end
 
+        Entity(veh).state:set('initVehicle', true, true)
+        Entity(veh).state:set('setVehicleProperties', props, true)
+
+        lib.waitFor(function()
+            local state = Entity(veh).state
```
Manason, line 264: cache the state outside the waitFor and use your cached state for setting the state bags as well

Manason (approved):

modules/lib.lua
solareon: Would it still update the value of the statebag? We are trying to see if it cleared.

Manason: Yes, this should be a reference to the statebag that's localized, I believe the memory location is shared.

SKITTLE6969:
this actually fixed the issue that I was having with vehicle properties going kaboom when pulling out a vehicle

Molicheu (approved):

PR #418, opened 2024-03-24
Add support for ox_lib Datadog and Grafana Loki logging

real-general-iroh:
## Description

This pull request adds an additional field to the logger function for tagging logs to be sent to ox_lib for API POST requests.  This allows people to have finer log control by being able to assign their own KV pairs to their logs, further enabling searching and querying.  See [lib.logger](https://overextended.dev/ox_lib/Modules/Logger/Server#liblogger) documentation in regards to the `vararg`.

## Checklist

### DIFF
--- a/modules/logger.lua
+++ b/modules/logger.lua
@@ -137,7 +137,7 @@ end
 ---@field message string the message attached to the log
 ---@field webhook? string Discord logs only. url of the webhook this log should send to
 ---@field color? string Discord logs only. what color the message should be
----@field tags? string[] Discord logs only. tags in discord. Example: {'<@%roleid>', '@everyone'}
+---@field tags? string[] Discord logs or ox_lib logs. tags in discord. Example: {'<@%roleid>', '@everyone'}. ox_lib logs: tags for datadog and grafana loki
 
 ---Logs using ox_lib if ox_lib logging is configured. Additionally logs to discord if a web hook is passed.
 ---@param log Log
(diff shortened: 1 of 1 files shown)


solareon:
So this shouldn't reuse the discord tags. That is used for if you desire to ping a role or person from a discord webhook. It should be added as a new field

zeiglermensch:
> So this shouldn't reuse the discord tags. That is used for if you desire to ping a role or person from a discord webhook. It should be added as a new field

I was thinking the same thing.  I will go ahead and make that change, now.

solareon (changes requested):

modules/logger.lua
```diff
@@ -138,6 +138,7 @@ end
 ---@field webhook? string Discord logs only. url of the webhook this log should send to
 ---@field color? string Discord logs only. what color the message should be
 ---@field tags? string[] Discord logs only. tags in discord. Example: {'<@%roleid>', '@everyone'}
+---@field oxLibTags? string[] -- Tags for ox_lib logger
```
solareon, line 141:
```suggestion
---@field oxLibTags? string -- Tags for ox_lib logger
```

Testing this it looks like it needs to be just a flat string with : and , to separate kvps. Passing a table just gives table 0x<whatever> in the log. Unless you have an example of how this can be passed as a table.

solareon (approved):

modules/logger.lua
solareon: @real-general-iroh do you have an example of this working with a table?

PR #434, opened 2024-04-08
add enableCityHall To view jobs within CityHall

bicatcho11:
## Description

add enableCityHall To view jobs within CityHall

## Checklist

### DIFF
--- a/shared/jobs.lua
+++ b/shared/jobs.lua
@@ -1,351 +1,369 @@
 ---@type table<string, Job>
 return {
-	['unemployed'] = {
-		label = 'Civilian',
-		defaultDuty = true,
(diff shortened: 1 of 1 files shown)


Manason:
Config like this should be in city hall resource.

bicatcho11:
> Config like this should be in city hall resource.

@Manason 

Do you have any suggestions for sorting jobs into two types?
For example, governmental and civil?
We can come up with a better result than the closure

Manason:
> > Config like this should be in city hall resource.
> 
> @Manason
> 
> Do you have any suggestions for sorting jobs into two types? For example, governmental and civil? We can come up with a better result than the closure

Do so in the city hall resource.

PR #436, opened 2024-04-10
feat(modules/lib): add veh spawn routing buckets

### LINKED ISSUE #435: Vehicle spawning in buckets does not work
### Summary

Spawning a car outside of the main bucket (outside of bucket 0) does not work

### Reproduction

1. Spawn a vehicle in any bucket besides bucket 0
2. Vehicle will not spawn
3. Set yourself in the main (0) bucket
4. Vehicle spawns as soon as you enter bucket 0

### Expected behavior

It's expected that a vehicle will spawn in your bucket

### Actual behavior

Does not spawn and gets "preloaded" for when you return to bucket 0

### Additional context

This used to work, then broke at some point in qbox development

### Current Version

current

### Custom Resources

N/A

solareon:
vehicle server side spawns aren't routing bucket aware. this adds two options for setting the routing bucket of the resultant spawned vehicle before returning.

## Description

fixes #435 by adding options to specify routing bucket in the params or via obtaining it from the warped in ped 

## Checklist


Manason:
Definitely needs to be tested. Last time we tried routing bucket things, natives were broken.

solareon:
> Definitely needs to be tested. Last time we tried routing bucket things, natives were broken.

Oh for sure. Didn't have time to test but figured albert could try it

modules/lib.lua
```diff
@@ -268,6 +270,11 @@ if isServer then
 
         if ped then
             SetPedIntoVehicle(ped, veh, -1)
+            bucket = GetEntityRoutingBucket(ped) or nil
+        end
+
+        if bucket then
+            SetEntityRoutingBucket(veh, bucket)
```
FjamZoo, line 277: Add an additional check to make sure that the bucket is not "0" no reason to set the bucket to 0 when its by default spawned at 0.

also the native has been fixed using oal

solareon:
```suggestion
            bucket = GetEntityRoutingBucket(ped) or nil
        end

        if bucket and bucket > 0 then
            SetEntityRoutingBucket(veh, bucket)
```

modules/lib.lua
```diff
@@ -235,6 +236,7 @@ if isServer then
         local warp = params.warp
         local ped = type(warp) == 'number' and warp or (sourceType == 'number' and warp and source or nil)
         local props = params.props
+        local bucket = params.bucket or nil
```
FjamZoo, line 239: why are we doing a "or nil"?

its either params.bucker or nil anyway, no reason to specify it

solareon:
```suggestion
        local bucket = params.bucket
```
because reasons

BerkieBb (approved):

mafewtm (approved):
Tested and works without issue as far as I can tell

PR #438, opened 2024-04-11
fix(server/player): do not zero metadata on logout

solareon:
statebags shouldn't be nil on logout but maybe if you got potatoes connected to potatoes then you'll be a nil bagholder

## Description

Statebags shouldn't be nil that fast on logout but in case they are don't zero the players metadata needs.

## Checklist


Randolio (approved):

jnccloud:
Is this why I have random players who wake up dead?

PR #439, opened 2024-04-11
fix(modules/lib): spawn temp vehicle under map

solareon:
## Description

When spawning vehicles serverside we need to create a temporary vehicle to get the vehicle type. For most vehicles this is okay but some can explode when being spawned at 0,0,0. this pushes the spawn under the map so that they don't go boom

## Checklist


TonybynMp4 (approved):
Hum sure why not 🤔

Randolio (approved):

PR #440, opened 2024-04-13
fix(server/player): only set state when possible

solareon:
## Description

Trying to set a statebag when the player is offline probably ain't gonna work. This moves these set events into a conditional that should correct issues with offline player data management

## Checklist


solareon:
Fixes https://github.com/Qbox-project/qbx_management/issues/56

Manason (approved):
What about setMetadata?

Also could move the var creation into the if source statement

solareon:
> What about setMetadata?
> 
> Also could move the var creation into the if source statement

good catch. Was looking for the localized state and missed this one

mafewtm (approved):

Manason (approved):

PR #442, opened 2024-04-15
feat(modules/lib): cardinal direction locales

mafewtm:
## Description

Added locales for the `qbx.getCardinalDirection` returns

## Checklist


Manason (commented):
Does this work since lib is a module? I wouldn't expect it to work since the locales are exported with the module

PR #445, opened 2024-04-17
refactor(client/server): Reintroduce Bucketing and Release streamingRequest

googleoblivion:
## Description

This adds back in bucketing to internal multicharacter, creating a callback to change player bucket. Refactors the use of bucket native to use internal SetPlayerBucket function. Finally releases a model request not being released after use.

#226

## Checklist

### DIFF
--- a/server/character.lua
+++ b/server/character.lua
@@ -47,7 +47,7 @@ lib.callback.register('qbx_core:server:loadCharacter', function(source, citizenI
     local success = Login(source, citizenId)
     if not success then return end
 
-    SetPlayerRoutingBucket(source, 0)
+    SetPlayerBucket(source, 0)
     logger.log({
         source = 'qbx_core',
         webhook = config.logging.webhook['joinleave'],
--- a/server/character.lua
+++ b/server/character.lua
@@ -76,9 +76,9 @@ lib.callback.register('qbx_core:server:createCharacter', function(source, data)
     return newData
 end)
 
-lib.callback.register('qbx_core:server:characterBucket', function (source)
+lib.callback.register('qbx_core:server:setCharBucket', function(source)
     SetPlayerBucket(source, source)
-    return GetPlayerRoutingBucket(source) == source
+    assert(GetPlayerRoutingBucket(source) == source, 'Multicharacter bucket not set.')
 end)
 
 RegisterNetEvent('qbx_core:server:deleteCharacter', function(citizenId)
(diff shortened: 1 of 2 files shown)


mafewtm (approved):

server/character.lua
```diff
@@ -69,13 +69,18 @@ lib.callback.register('qbx_core:server:createCharacter', function(source, data)
...
+        SetPlayerBucket(source, 0)
     end
 
     lib.print.info(('%s has created a character'):format(GetPlayerName(source)))
     return newData
 end)
 
+lib.callback.register('qbx_core:server:characterBucket', function (source)
+    SetPlayerBucket(source, source)
+    return GetPlayerRoutingBucket(source) == source
```
Manason, line 81: Under what scenario would setting the bucket fail?

server/character.lua
```diff
@@ -69,13 +69,18 @@ lib.callback.register('qbx_core:server:createCharacter', function(source, data)
...
     if GetResourceState('qbx_spawn') == 'missing' then
-        SetPlayerRoutingBucket(source, 0)
+        SetPlayerBucket(source, 0)
     end
 
     lib.print.info(('%s has created a character'):format(GetPlayerName(source)))
     return newData
 end)
 
+lib.callback.register('qbx_core:server:characterBucket', function (source)
```
Manason, line 79: Rename with a leading verb

Maybe 'qbx_core:server:setCharBucket'

Manason (changes requested):

server/character.lua
googleoblivion: Would it be better as an event rather than a callback in that case?

Manason: I think the callback is fine. Callbacks don't have to return anything, but you are doing logic in response to the return. If this is just a safety check, then assert it here rather than return.

googleoblivion: @Manason i used callback.await, so isn't the return necessary?

Manason: I believe if there is no return specified, a default return of nil will be present.

googleoblivion: I can revert before latest commits and go back to .await then, just return true regardless and assert error for the bucket check.

Manason:
Can you double check that this doesn't assert in-game? Last time we had issues with the native, but I believe a subsequent server artifact corrected those.

mafewtm:
Once its checked we can merge this

FjamZoo:
> Can you double check that this doesn't assert in-game? Last time we had issues with the native, but I believe a subsequent server artifact corrected those.

yes its been fixed

PR #450, opened 2024-04-20
Fix #443 - Fix for default duty not working if set to false

### LINKED ISSUE #443: [Possible Bug] Duty Status is Setting Incorrectly at Login (Not defaulting to what the shared/jobs.lua has in place)
### Summary

Duty status is not defaulting to what the shared/jobs.lua has.

### Reproduction

1. Make sure to go on/off duty (Make sure duty is true/false). Preferrably go on duty to make it easier to identify.
2. After going on/off duty, logoff (F8 Quit).
3. Log back on. After loading, check your duty status.
4. It should be set to whatever your last duty status is.

### Expected behavior

Since I have defaultDuty set to false within shared/jobs.lua it should be forcing the player when they login to off-duty. Yes, I have ForceJobDefaultDutyAtLogin set to true.

### Actual behavior

After updating from 1.6.1 qbx_core, players will login on duty when the defaultDuty status is set to false in shared/jobs.lua. Within the shared/main.lua ForceJobDefaultDutyAtLogin is set to true.

### Additional context

N/A

### Current Version

v1.10.1

### Custom Resources

Nothing that would directly effect this issue. It only happened after an upgrade from 1.6.1.

Scullyy:
## Description

Fix for default duty not working if set to false

## Checklist

### DIFF
--- a/server/player.lua
+++ b/server/player.lua
@@ -361,12 +361,15 @@ function CheckPlayerData(source, playerData)
     local job = GetJob(playerData.job?.name) or GetJob('unemployed')
     assert(job ~= nil, 'Unemployed job not found. Does it exist in shared/jobs.lua?')
     local jobGrade = GetJob(playerData.job?.name) and playerData.job.grade.level or 0
+    if QBX.Shared.ForceJobDefaultDutyAtLogin and (job.defaultDuty ~= nil) then
+        playerData.job.onduty = job.defaultDuty
+    end
     playerData.job = {
         name = playerData.job?.name or 'unemployed',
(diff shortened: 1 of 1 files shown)


mafewtm (approved):

Manason (commented):
This logic looks the same, what am I missing?

Scullyy:
> This logic looks the same, what am I missing?

in the current logic if job.defaultDuty was set to false then it would move onto playerData.job?.onduty

Manason (approved):

PR #451, opened 2024-04-20
Replace QB items.lua and weapons.lua with Ox

Scullyy:
## Description
This replaces the QBCore items.lua and weapons.lua with the ox one, I also had to add a WeaponHashes table for reverse hashing since ox uses strings. I have open PR's on other resources to be compatible with this also.

## Checklist


Manason (changes requested):
This is a breaking change. The data format of ox_inventory is not the same as shared/weapons.lua

TonybynMp4 (approved):
YES I LOVE YOU

Scullyy:
The items.lua will need to replace the one inside ox_inventory with the recipe. If not using the oldItems method this may need it's own bridge inside ox_inventory.

BerkieBb:
I am not sure about this, I would rather not have an items.lua anymore in the future as they all get picked up by ox_inventory and placed in the data/items.lua anyway, otherwise they don't work

SKITTLE6969:
only issue that i see is when ox inventory try to convert the item lua in shared what happens then

Issue #457, opened 2024-04-23
hunger and thirst states desynced from metadata


Demigod916:
### Summary

hunger and thirst states desynced from metadata

### Reproduction

1. log in
2. have a script affect your hunger/thirst through setMetadata
3. wait
4. die

### Expected behavior

state and metadata synced for backwards compat.

### Actual behavior

dies while hunger and thirst are not empty on hud using metadata

### Additional context

assume that this can be fixed with a onMetaChange handler in the core

### Current Version

v1.11.0

### Custom Resources

.

Demigod916:
fix manason pls

Demigod916:
```lua
AddEventHandler('qbx_core:server:onSetMetaData', function(key, _, new, source)
    if key == 'hunger' or key == 'thirst' or key == 'stress' then
        Player(source).state:set(key, new, true)
    end
end)
```

Demigod916:
nvm looks like the state is being synced in the self.Functions.SetMetaData. but still experiencing the desynced values between metadata and state for status. will work oin

Demigod916:
```lua
local PlayerState = LocalPlayer.state

CreateThread(function ()
    while true do
        print("ON LOOP")
        lib.print.info("STATE:", "HUNGER:", PlayerState.hunger)
        lib.print.info("STATE:", "THIRST:", PlayerState.thirst)

        lib.print.info("META:", "HUNGER:", QBX.PlayerData.metadata.hunger)
        lib.print.info("META:", "THIRST:", QBX.PlayerData.metadata.thirst)
        Wait(500)
        print("========================================================")
    end
end)


RegisterCommand('addhunger', function ()
    TriggerServerEvent('consumables:server:addHunger', QBX.PlayerData.metadata.hunger + 1000) --- this event just sets the hunger using setMetaData function
end)

RegisterCommand('addthirst', function ()
    TriggerServerEvent('consumables:server:addThirst', QBX.PlayerData.metadata.thirst + 1000)--- this event just sets the thirst using setMetaData function
end)


RegisterNetEvent('hud:client:UpdateNeeds', function (hunger, thirst)
    print("ON UPDATE")
    lib.print.info("STATE:", "HUNGER:", PlayerState.hunger)
    lib.print.info("STATE:", "THIRST:", PlayerState.thirst)

    lib.print.info("META:", "HUNGER:", QBX.PlayerData.metadata.hunger)
[...]

PR #459, opened 2024-04-24
fix(server/player): status desync

Demigod916:
## Description

This fixes an issue where the statebag statuses were clamped to 0-100 while the meta statuses wcould go over 100. causing them to become desynced. this would cause huds that use metadata values to say that players had more hunger/thirst than they really had and they would start dying for no apparent reason. it will now clamp both to 100.

repro here in issue #457 

## Checklist

- [ x] My pull request fits the contribution guidelines & code conventions.

### DIFF
--- a/server/player.lua
+++ b/server/player.lua
@@ -530,7 +530,8 @@ function CreatePlayer(playerData, Offline)
     function self.Functions.SetMetaData(meta, val)
         if not meta or type(meta) ~= 'string' then return end
         if (meta == 'hunger' or meta == 'thirst' or meta == 'stress') and self.PlayerData.source then
-            Player(self.PlayerData.source).state:set(meta, lib.math.clamp(val, 0, 100), true)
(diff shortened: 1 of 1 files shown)


Manason:
Which is correct? Are these values supposed to be clamped to 100?

Demigod916:
> Which is correct? Are these values supposed to be clamped to 100?

no clue, but it doesnt really make sense for hunger and thirst to be infinite? maybe a config for max cap. defaulted to 100 is goor

Manason (approved):

Demigod916:
> Which is correct? Are these values supposed to be clamped to 100?

https://github.com/qbcore-framework/qb-core/blob/5c60c94197d0a31be30ca7f2f06c4a894699f2bd/server/player.lua#L294

looks like clamped to 100 is the intended behavior in qb-core

Randolio (approved):

PR #462, opened 2024-04-27
feat(modules/lib): getVehicleModName and getVehicleLiveryName

mafewtm:
## Description

Adds a function to get the vehicle mod name and vehicle livery name. Came across a use for it and realized we didn't have it in our lib so I figured I'd toss it in here. Not sure what everyone else's options on it are. 

## Checklist

### DIFF
--- a/modules/lib.lua
+++ b/modules/lib.lua
@@ -395,6 +395,23 @@ else
         return GetLabelText(GetMakeNameFromVehicleModel(GetEntityModel(vehicle)))
     end
 
+    ---Returns the mod type name of the given vehicle.
+    ---@param vehicle integer
+    ---@param modType integer
+    ---@param modIndex integer
+    ---@return string
+    function qbx.getVehicleModName(vehicle, modType, modIndex)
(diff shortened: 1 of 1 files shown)


KostaZx (approved):

modules/lib.lua
```diff
@@ -395,6 +395,23 @@ else
         return GetLabelText(GetMakeNameFromVehicleModel(GetEntityModel(vehicle)))
     end
 
+    ---Returns the mod type name of the given vehicle.
+    ---@param vehicle integer
+    ---@param modType integer
+    ---@param modIndex integer
+    ---@return string
```
Manason, line 402: Are there external documentation links that might be helpful here? I'm not sure what a mod type nor mod index are

Manason (approved):

modules/lib.lua
TonybynMp4: Mod type & mod index are very self explanatory,
type can be the transmission, turbo, exhaust etc
and index is.. the index in that type

PR #469, opened 2024-05-13
feat: handle caller errors using errors as values

Manason:
When the caller of an API function passes an illegal value (equivalent to a 4xx HTTP error), changing existing assertions to instead return success, errorResult to allow the caller to handle those errors or else fail silently. This is a departure from assertions, which require errors to be caught and handled or else blow up the stack. The thinking is that this default behavior of silently failing is what callers would prefer.

assertions continued to be used for types of errors that clients should not be handling, but are rather indicative of a bug. These do not have an error code as they are unexpected faults in the program and therefore should only be handled for purposes like logging, which do not require differentiation among the causes of the assertion violation.

Also introduced a types.lua to hold general types which don't really belong anywhere else. We can consider moving all types here in a future PR.

### DIFF
--- a/server/player.lua
+++ b/server/player.lua
@@ -114,6 +114,7 @@ function SetPlayerPrimaryJob(citizenid, jobName)
 
     player.PlayerData.job = toPlayerJob(jobName, job, grade)
     player.Functions.Save()
+
     if not player.Offline then
         player.Functions.UpdatePlayerData()
         TriggerEvent('QBCore:Server:OnJobUpdate', player.PlayerData.source, player.PlayerData.job)
@@ -139,6 +140,7 @@ function AddPlayerToJob(citizenid, jobName, grade)
             message = 'players cannot be added to the unemployed job'
         }
     end
+
     local job = GetJob(jobName)
     if not job then
         return false, {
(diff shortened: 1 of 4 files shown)


server/player.lua
```diff
@@ -109,17 +129,46 @@ exports('SetPlayerPrimaryJob', SetPlayerPrimaryJob)
...
+    if not player then
+        return false, {
+            code = 'player_not_found',
+            message = string.format('player not found with citizenid %s', citizenid)
+        }
+    end
+    if player.PlayerData.jobs[jobName] == grade then
+        return true
+    end
+    if qbx.table.size(player.PlayerData.jobs) >= maxJobsPerPlayer and not player.PlayerData.jobs[jobName] then
```
BerkieBb, line 166: question, why not add spaces between if statements? I personally don't like it when they're grouped together as they're meant to do different things and may make it difficult to read.

BerkieBb (approved):

server/player.lua
Manason: added more whitespace.

KostaZx (approved):

PR #477, opened 2024-06-05
feat(bridge/qb/server) assertions for deprecated qb-inventory functions

mafewtm:
## Description

Assert that qbx_core will not work with qb-inventory and that the user should utilize ox_inventory instead.

## Checklist

### DIFF
--- a/server/player.lua
+++ b/server/player.lua
@@ -538,8 +538,6 @@ function CheckPlayerData(source, playerData)
     }
     playerData.gangs = gangs or {}
     playerData.position = playerData.position or defaultSpawn
-
-    assert(GetResourceState('qb-inventory') ~= 'started', 'qb-inventory is not compatible with qbx_core. use ox_inventory instead')
     playerData.items = {}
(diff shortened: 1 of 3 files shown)


server/player.lua
```diff
@@ -538,7 +538,9 @@ function CheckPlayerData(source, playerData)
     }
     playerData.gangs = gangs or {}
     playerData.position = playerData.position or defaultSpawn
-    playerData.items = GetResourceState('qb-inventory') ~= 'missing' and exports['qb-inventory']:LoadInventory(playerData.source, playerData.citizenid) or {}
+
+    assert(GetResourceState('qb-inventory') ~= 'started', 'qb-inventory is not compatible with qbx_core. use ox_inventory instead')
```
solareon, line 542:
```suggestion
```
If we are going to just provide an empty table then we don't need to assert here.

solareon (approved):

server/player.lua
mafewtm: You're right. I'll fix that

Manason (approved):

PR #478, opened 2024-06-05
chore: remove leftover data tables

t0r3tto:
## Description

remove deprecated character data tables that are no longer used
image
(^ clean recipe deployment warnings)

## Checklist

* [x]  I have personally loaded this code into an updated Qbox project and checked all of its functionality.
* [X]  My pull request fits the contribution guidelines & code conventions.


solareon:
Sadly these have to stay. If users are converting from `qb-core` then they may have characters in those tables that need to be cleaned up when a character is deleted. A possible solution for this would be to copy a different config file in the txadmin recipe for clean installs over this file to prevent the warnings.

solareon:
We are also looking at possibly moving these to convars which would fix the issue as well as the recipe would include on what is needed.

solareon (approved):

Manason (approved):
After discussion, I think the right choice is indeed to cater to recipe installs. That means that those converting from qb-core will have to pay special attention to the config to ensure they add any missing tables that they have. They should already be going through the config as part of the conversion process however, so this seems like the right call.

PR #473, opened 2024-05-22
refactor(character): Remove callback and use tutorial native.

googleoblivion:
## Description

Linden said bad.

## Checklist

- [ ] I have personally loaded this code into an updated Qbox project and checked all of its functionality.
- [x] My pull request fits the contribution guidelines & code conventions.

PR #480, opened 2024-06-05
refactor(character): Use tutorial natives instead of manual bucket setting

solareon:
## Description

Routing buckets may overlap. The natives do a better job of hiding the characters. This is similar to #473 but moves the tutorial end to the `QBCore:Client:OnPlayerLoaded` event to end the tutorial session with the appropriate native.

Additionally it only calls the tutorial session start one time if somehow the choosecharacter() function got called twice without being disabled cause who knows if that works right.

NOTE: I have **NOT** tested this. Do not merge without extensive testing in a multiplayer environment. 

## Checklist


Molicheu (approved):

Manason (changes requested):
Blocking on testing to be confirmed (in a multi-client environment)

mafewtm:
Tested by @D4isDAVID and appears to work as expected

Manason (approved):

PR #482, opened 2024-06-06
feat(lib): return oldValue with handler

solareon:
## Description

Improves functionality of the entityStateHandler by returning the previous statebag value to allow for easier comparison.

## Checklist


Manason (changes requested):
I think you can get the current value by checking the state bag. Will approve if someone tests and confirms this is not the case.

PR #488, opened 2024-06-18
fix(bridge/qb/shared/main): context check for requiring server file

Demigod916:
since this is a shared file, the client copy of the file errors out immediately on script start since its trying to require a file thats not available to the client.

### DIFF
--- a/bridge/qb/shared/main.lua
+++ b/bridge/qb/shared/main.lua
@@ -1,6 +1,6 @@
 local qbShared = require 'shared.main'
 
-if lib.context == 'server' then
+if IsDuplicityVersion() then
     local starterItems = require 'config.server'.starterItems
     ---@deprecated use starterItems in config/server.lua
     qbShared.StarterItems = {}
--- a/bridge/qb/shared/main.lua
+++ b/bridge/qb/shared/main.lua
@@ -1,16 +1,17 @@
 local qbShared = require 'shared.main'
 
-local starterItems = require 'config.shared'.starterItems or {}
----@deprecated use starterItems in config/server.lua
+local starterItems = require 'config.shared'.starterItems
+---@deprecated use starterItems in config/shared.lua
 qbShared.StarterItems = {}
 
-assert(next(starterItems), 'starterItems is missing or incorrect in shared/config.lua')
-for i = 1, #starterItems do
-    local item = starterItems[i]
-    qbShared.StarterItems[item.name] = {
-        amount = item.amount,
-        item = item.name,
-    }
+if type(starterItems) == 'table' then
+    for i = 1, #starterItems do
+        local item = starterItems[i]
+        qbShared.StarterItems[item.name] = {
+            amount = item.amount,
+            item = item.name,
+        }
+    end
 end
 
 ---@deprecated use lib.math.groupdigits from ox_lib
--- a/bridge/qb/shared/main.lua
+++ b/bridge/qb/shared/main.lua
@@ -1,15 +1,17 @@
-local starterItems = require 'config.server'.starterItems
 local qbShared = require 'shared.main'
 
----@deprecated use starterItems in config/server.lua
-qbShared.StarterItems = {}
+if lib.context == 'server' then
+    local starterItems = require 'config.server'.starterItems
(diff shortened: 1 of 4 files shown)


bridge/qb/shared/main.lua
```diff
@@ -1,15 +1,17 @@
-local starterItems = require 'config.server'.starterItems
 local qbShared = require 'shared.main'
 
----@deprecated use starterItems in config/server.lua
-qbShared.StarterItems = {}
+if lib.context == 'server' then
```
solareon, line 3:
```suggestion
if IsDuplicityVersion() then
```
I believe this is functionally the same. I've not seen lib.context used and it's not specifically mentioned in the ox_lib docs so it may be intended for internal use only.

solareon (changes requested):

bridge/qb/shared/main.lua
```diff
@@ -1,15 +1,17 @@
-local starterItems = require 'config.server'.starterItems
 local qbShared = require 'shared.main'
 
----@deprecated use starterItems in config/server.lua
-qbShared.StarterItems = {}
```
Manason, line 5: Keep the default for client table? Otherwise this would now be nil when previously was an empty table.

Manason (changes requested):

bridge/qb/shared/main.lua
solareon: Yeah, might be better to just move the starter items to shared since the client is going to get them anyway at start it's not a big deal for the client to know what they are

bridge/qb/shared/main.lua
```diff
@@ -1,9 +1,10 @@
-local starterItems = require 'config.server'.starterItems
 local qbShared = require 'shared.main'
 
+local starterItems = require 'config.shared'.starterItems or {}
 ---@deprecated use starterItems in config/server.lua
 qbShared.StarterItems = {}
 
+assert(next(starterItems), 'starterItems is missing or incorrect in shared/config.lua')
```
Manason, line 7: Is empty starter items not a valid entry?

Manason (changes requested):

bridge/qb/shared/main.lua
solareon: non existent starterItems isn't valid cause the loop will fail. Since we moved it from server to shared all the existing configurations will be broken

Manason: It defaults to empty table in the above line if nil

config/shared.lua
```diff
@@ -1,5 +1,20 @@
...
+                assert(GetResourceState('qbx_idcard') == 'started', 'qbx_idcard resource not found. Required to give an id_card as a starting item')
+                return exports.qbx_idcard:GetMetaLicense(source, {'id_card'})
+            end
+        },
+        { name = 'driver_license', amount = 1, metadata = function(source)
+                assert(GetResourceState('qbx_idcard') == 'started', 'qbx_idcard resource not found. Required to give an id_card as a starting item')
+                return exports.qbx_idcard:GetMetaLicense(source, {'driver_license'})
+            end
+        },
+    },
```
solareon, line 19:
```suggestion
    starterItems = function()
        ---@type { name: string, amount: integer, metadata: fun(source: number): table }[]
        local starterItemTable = { -- Character starting items
            { name = 'phone', amount = 1 },
            { name = 'id_card', amount = 1, metadata = function(source)
                    assert(GetResourceState('qbx_idcard') == 'started', 'qbx_idcard resource not found. Required to give an id_card as a starting item')
                    return exports.qbx_idcard:GetMetaLicense(source, {'id_card'})
                end
            },
            { name = 'driver_license', amount = 1, metadata = function(source)
                    assert(GetResourceState('qbx_idcard') == 'started', 'qbx_idcard resource not found. Required to give an id_card as a starting item')
                    return exports.qbx_idcard:GetMetaLicense(source, {'driver_license'})
                end
            },
        }
        assert(next(starterItemTable), 'starterItems is missing or incorrect in shared/config.lua')
        return starterItemTable
    end
```

@Manason this should be correct

Manason (approved):

PR #497, opened 2024-06-24
feat: HasGroup export

solareon:
## Description

This adds a utility export that allows for a string/table of job/gang/citizenId to be passed against a function and quickly return if the condition matches. Utilizes ox_target code for group/citizenId matching.

## Checklist

### DIFF
--- a/client/functions.lua
+++ b/client/functions.lua
@@ -43,3 +43,72 @@ function GetPlayerData()
 end
 
 exports('GetPlayerData', GetPlayerData)
+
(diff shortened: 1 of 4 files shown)


Manason (changes requested):
Could we extract shared code to a shared function? Nice idea moving this to core to be generally available. Also, maybe we should rename to make it clear that we're only considering the primary group? Maybe HasPrimaryGroup?

mafewtm (approved):

PR #500, opened 2024-06-27
"No database selected" fix

Molicheu:
## Description

This pull request adds a conditional statement to create the qboxproject database only if it does not already exist. This ensures that existing data is not overwritten unintentionally, enhancing data safety during setup and updates.
## Checklist

### DIFF
--- a/qbx_core.sql
+++ b/qbx_core.sql
@@ -1,3 +1,7 @@
+CREATE DATABASE IF NOT EXISTS `qboxproject`;
+
+USE `qboxproject`;
+
 CREATE TABLE IF NOT EXISTS `players` (
   `id` int(11) NOT NULL AUTO_INCREMENT,
   `citizenid` varchar(50) NOT NULL COLLATE utf8mb4_unicode_ci,


Manason (changes requested):
Doesn't txAdmin create the database by default for you? Will it be named qboxproject? Is there a way to only use 'qboxproject' if it doesn't exist?

BigSmoKe07:
Certainly! Here's a more professional version of your message:

---

When performing a manual installation, it is necessary to select a database. However, if you are using txAdmin's deployer, it will manage the database connection for you. For reference, please see the txAdmin documentation [here](https://github.com/tabarra/txAdmin/blob/216e28a545875af232c605cbb181044039b0ff91/docs/recipe.md?plain=1#L228). Additionally, if the SQL query encounters existing tables, they will be skipped without affecting the current setup.

Manason:
So this PR is only for those who aren't using txAdmin and will create a database for them and then select it? What if they already have a database created, then this will populate the wrong one?

Molicheu:
> Doesn't txAdmin create the database by default for you? Will it be named qboxproject? Is there a way to only use 'qboxproject' if it doesn't exist?

Yes it does, I was having issues when I only had qbx_core 

> So this PR is only for those who aren't using txAdmin and will create a database for them and then select it? What if they already have a database created, then this will populate the wrong one?

Yes to the first question and they can just change the name of the database in the SQL to the one they're using

Manason:
> > So this PR is only for those who aren't using txAdmin and will create a database for them and then select it? What if they already have a database created, then this will populate the wrong one?
> 
> Yes to the first question and they can just change the name of the database in the SQL to the one they're using

My issue is that this would fail silently. i.e. the SQL script would run, but the user would not have their database populated as they expect.

PR #502, opened 2024-07-01
fix(lib): spawn vehicle property setting when warping

### LINKED ISSUE #501: Spawning vehicle with warp can fail to set vehicle properties
### Summary

When the warping client isn't the entity owner, the vehicle properties seem to fail to set

### Reproduction

1. Have two clients in close proximity
2. run spawnrepo a few times until the license plate of the spawned vehicle is not 'TEST'
```lua
RegisterCommand('spawnrepro', function(source)
    qbx.spawnVehicle({ spawnSource = vec4(228.5577, -802.1774, 30.5724, 159.7563), model = 'adder', props = {plate = 'TEST'}, warp = GetPlayerPed(source)})
end)
```

### Expected behavior

The license plate of the spawned vehicle is 'TEST' since that's the property that was passed in.

### Actual behavior

The license plate has a chance of not being 'TEST'

### Additional context

This occurs when the vehicle entity owner is the client who is not running the spawnrepro command. I suspect there is a race condition that takes away ownership before the properties can be set, leading to default GTA properties instead, although I haven't been able to nail down a root cause.

### Current Version

v1.15.0

### Custom Resources

N/A

Manason:
Fixes #501 

- Tested
- Re-ordered operations so that first the vehicle bucket is set, then the player is warped, and then we wait for the vehicle to have an owner (warp is set it wait for the warping player to become the owner) and then the properties are set.
- Ensures that properties are set, or else deletes the vehicle and throws an error.

### DIFF
--- a/modules/lib.lua
+++ b/modules/lib.lua
@@ -236,7 +236,7 @@ if isServer then
         local warp = params.warp
         local ped = type(warp) == 'number' and warp or (sourceType == 'number' and warp and source or nil)
         local props = params.props
-        local bucket = params.bucket
+        local bucket = params.bucket or ped and GetEntityRoutingBucket(ped) or nil
 
         ---@type vector4
         local coords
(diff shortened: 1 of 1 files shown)


modules/lib.lua
```diff
@@ -259,23 +259,40 @@ if isServer then
...
+
+        if props and type(props) == 'table' and props.plate then
+            state:set('setVehicleProperties', props, true)
+            local success = lib.waitFor(function()
+                if qbx.string.trim(GetVehicleNumberPlateText(veh)) == qbx.string.trim(props.plate) then
+                    return true
+                end
+            end, 'Failed to set vehicle properties within 5 seconds', 5000)
+            if not success then
+                DeleteEntity(veh)
```
solareon, line 291: A nice touch.

solareon (approved):
Makes sense, the order of operations between warp/prop set wasn't really discussed and this from a flow/onesync pov makes more sense.

Molicheu (approved):

PR #503, opened 2024-07-03
add server side playerLoggedOut Event Handler

ARSSANTO:
## Description

This event will work as handler  like 

```
-- Add any additional event handlers for player logout
-- Server Side
AddEventHandler('qbx_core:server:playerLoggedOut', function(source)
    -- Add any additional actions you want to perform when a player logs out
    print('Player with source ID ' .. source .. ' has logged out.')
end)
```

## Checklist
- [yes ] I have personally loaded this code into an updated Qbox project and checked all of its functionality.
- [ yes] My pull request fits the contribution guidelines & code conventions.

### DIFF
--- a/server/player.lua
+++ b/server/player.lua
@@ -562,6 +562,7 @@ function Logout(source)
     QBX.Players[source] = nil
     GlobalState.PlayerCount -= 1
     TriggerClientEvent('qbx_core:client:playerLoggedOut', source)
+    TriggerEvent('qbx_core:server:playerLoggedOut', source)
 end
 
 exports('Logout', Logout)


Manason (commented):
What's the use case for this? Do you have a specific need in mind?

ARSSANTO:
> What's the use case for this? Do you have a specific need in mind?

Yes I have and already using it so on the server side there if this event triggers I can send logs then also I can delete props from logged out player  very easily and many more server side action can be taken directly while player logged out .

rather going to client side and back to server side 

```lua
AddEventHandler('qbx_core:server:playerLoggedOut', function(source)
    -- Add any additional actions you want to perform when a player logs out
    print('Player with source ID ' .. source .. ' has logged out.')
end)
```

Manason (approved):

FjamZoo:
> What's the use case for this? Do you have a specific need in mind?

Its a good addition makes it so we dont need to use the standard "onPlayerDrop" event while also supporting logging out of multi characters.

FjamZoo (approved):

PR #504, opened 2024-07-04
fix: propagate errors for job/gang commands

Manason:
Asserting that all the commands for job/gang modification are successful so that the error is printed in console if it fails


solareon (approved):
Guess I never hit approve earlier

mafewtm (approved):

PR #506, opened 2024-07-05
replaced lib.waitFor with while, in qbx.spawnVehicle

abdel1touimi:
## Checklist

### DIFF
--- a/modules/lib.lua
+++ b/modules/lib.lua
@@ -267,16 +267,22 @@ if isServer then
             SetPedIntoVehicle(ped, veh, -1)
         end
 
-        local foundOwner = lib.waitFor(function()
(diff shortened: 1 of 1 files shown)


solareon (changes requested):
This could also be improved by an extension to the lib.waitFor that provides the ability to return to a different callback, in this case to delete the entity rather than erroring.

abdel1touimi:
> This could also be improved by an extension to the lib.waitFor that provides the ability to return to a different callback, in this case to delete the entity rather than erroring.

Yes I agree. until then we have no choice here

Manason:
Alternative solution: https://github.com/Qbox-project/qbx_core/pull/507

PR #510, opened 2024-07-07
feat(client/events): add handler for triggering hud updates

solareon:
## Description

Adds statebag handlers to trigger hud updates for compatibility. Not required for qbx_hud but other huds that listen to the legacy NetEvents need these

## Checklist


Manason (approved):
Should this be in the bridge instead, or is it more appropriate here?

solareon:
> Should this be in the bridge instead, or is it more appropriate here?

good call, done.

mafewtm (approved):

Issue #463, opened 2024-04-29
Ban Checking


DevX32:
### Summary

using the export its bans an drop th player and saves into db but still the banned player can connect

### Reproduction

1. use the export exports.qbx_core:ExploitBan(playerId, origin)
2. and check it bans or not
3. if it bans go to db and check it saves or not
4. and if it saves and connect to server

### Expected behavior

do not let the banned player connect  the server

### Actual behavior

banned player can connect server

### Current Version

v1.11.1

### Custom Resources

none

KostaZx:
I don't see the point of the core having a ban system, just use txAdmin.

DevX32:
> I don't see the point of the core having a ban system, just use txAdmin.
not aying about ban system telling about ban checking

mafewtm:
Should be fixed as of 6ff85eb

PR #519, opened 2024-07-19
fix(server/functions): ban checking edge case

mafewtm:
## Description

There seems to be a weird edge case in ban checking which I believe is the issue in #463.

Currently we pull whichever license the player has (license2 first then license) and then check it against the bans table to see if the individual is banned. If you have an admin menu which bans by license but you have a license2, the check will give a false positive and allow the player through because it did not check by license. This fixes that by first checking by license2 and then by license to make sure we catch both cases.

Feel free to comment or start a discussion below. Idk if there's a better way we'd like to do this.

## Checklist

### DIFF
--- a/server/functions.lua
+++ b/server/functions.lua
@@ -332,10 +332,18 @@ exports('ToggleOptin', ToggleOptin)
 ---@return boolean
 ---@return string? playerMessage
 function IsPlayerBanned(source)
-    local plicense = GetPlayerIdentifierByType(source --[[@as string]], 'license2') or GetPlayerIdentifierByType(source --[[@as string]], 'license')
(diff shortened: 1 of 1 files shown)


mafewtm:
Now that I think about it, editing the fetchBan SQL is probably a better alternative than this. I'll do that tomorrow if y'all agree.

solareon:
> Now that I think about it, editing the fetchBan SQL is probably a better alternative than this. I'll do that tomorrow if y'all agree.

Yeah to use the same lookup as the fetch characters.

Manason (approved):

solareon (approved):

PR #520, opened 2024-07-19
fix(server/functions): exploitban yield until ids inserted

solareon:
## Description

Spawning off a thread for the ban insert while functionally correct the natives could fail to return the identifiers preventing the ban from being inserted. This moves the insert into a pcall which will yield until the ban is inserted.

## Checklist

### DIFF
--- a/server/functions.lua
+++ b/server/functions.lua
@@ -424,7 +424,7 @@ local function ExploitBan(playerId, origin)
         event = 'Anti-Cheat',
         color = 'red',
         tags = loggingConfig.role,
-        message = ('%s has been banned for exploiting %s'):format(name, origin)
+        message = success and ('%s has been kicked and banned for exploiting %s'):format(name, origin) or ('%s has been kicked for exploiting %s, ban insert failed'):format(name, origin)
     })
 end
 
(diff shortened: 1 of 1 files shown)


server/functions.lua
```diff
@@ -398,18 +398,17 @@ exports('GetCoreVersion', GetCoreVersion)
...
             discordId = GetPlayerIdentifierByType(playerId --[[@as string]], 'discord'),
             ip = GetPlayerIdentifierByType(playerId --[[@as string]], 'ip'),
             reason = origin,
             expiration = 2147483647,
             bannedBy = 'Anti Cheat'
-        })
-        assert(success, json.encode(errorResult))
-    end)
+        }))
+    if not success then lib.print.error(errorResult) end
```
Manason, line 411: Shouldn't it return here to prevent incorrectly logging that someone was banned and also if we fail to ban, maybe we want to fail the kick as well?

Manason (changes requested):

server/functions.lua
solareon: Well prior to this the ban would fail inserting anyway due to the natives returning nothing since the source was gone by the time we polled for licenses. I can see the benefit of returning after the kick and not logging since it didn't succeed

solareon: Refactored this with a better log message.

Manason (approved):

FjamZoo (approved):

PR #527, opened 2024-08-03
refactor(server/storage): block non-existent groups from PlayerData

### LINKED ISSUE #526: core does not validate that playergroups sql table contains valid jobs/gangs
### Summary

playergroups not validated before being added to playerdata

### Reproduction

Add gang/job to shared luas. Add gang/job to player via `/addjob` or `/addgang`. Remove gang/job from shared lua. The gang/job remains present in the `PlayerData.jobs` table.

### Expected behavior

Invalid jobs/gangs are removed from the `PlayerData.jobs` table either on server startup or player load and an error message is thrown to advise the user to remove this bad data from the sql table

### Actual behavior

gang/job remains present in the `PlayerData.jobs` table.

### Additional context

_No response_

### Current Version

v1.17.2

### Custom Resources

any multijob utilizing core exports.

solareon:
## Description

Adds cleanplayergroups command to remove invalid groups from the player_groups table to prevent issues with PlayerData loading. Additionally checks for group validity during player load to block invalid groups from being inserted into the respective PlayerData.jobs/gangs tables. Fixes #526

## Checklist

### DIFF
--- a/server/storage/players.lua
+++ b/server/storage/players.lua
@@ -325,7 +325,7 @@ local function fetchPlayerGroups(citizenid)
         local group = groups[i]
         if group.type == GroupType.JOB then
             jobs[group.group] = group.grade
-        else
+        elseif group.type == GroupType.GANG then
             gangs[group.group] = group.grade
         end
     end
--- a/server/storage/players.lua
+++ b/server/storage/players.lua
@@ -326,6 +326,8 @@ local function fetchPlayerGroups(citizenid)
         local validGroup = group.type == GroupType.JOB and GetJob(group.group) or GetGang(group.group)
         if not validGroup then
             lib.print.warn(('Invalid group %s found in player_groups table, Does it exist in shared/%ss.lua?'):format(group.group, group.type))
+        elseif not validGroup[group.grade] then
+            lib.print.warn(('Invalid grade %s found in player_groups table for %s %s, Does it exist in shared/%ss.lua?'):format(group.grade, group.type, group.group, group.type))
         elseif group.type == GroupType.JOB then
             jobs[group.group] = group.grade
         elseif group.type == GroupType.GANG then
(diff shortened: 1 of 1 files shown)


server/storage/players.lua
```diff
@@ -323,9 +323,12 @@ local function fetchPlayerGroups(citizenid)
...
         local group = groups[i]
-        if group.type == GroupType.JOB then
+        local validGroup = group.type == GroupType.JOB and GetJob(group.group) or GetGang(group.group)
+        if not validGroup then
+            lib.print.warn(('Invalid group %s found in player_groups table, Does it exist in shared/%ss.lua?'):format(group.group, group.type))
+        elseif group.type == GroupType.JOB then
             jobs[group.group] = group.grade
-        else
+        elseif group.type == GroupType.GANG then
             gangs[group.group] = group.grade
```
Manason, line 332: What is the group is valid, but the grade is not?

server/storage/players.lua
```diff
@@ -369,6 +372,24 @@ RegisterCommand('convertjobs', function(source)
...
+        local group = groups[i]
+        local validGroup = group.type == GroupType.JOB and GetJob(group.group) or GetGang(group.group)
+        if not validGroup then
+            MySQL.query.await('DELETE FROM player_groups WHERE `group` = ? AND type = ?', {group.group, group.type})
+            lib.print.info(('Remove invalid %s %s from player_groups table'):format(group.type, group.group))
+        end
+    end
+
+    lib.print.info('Removed invalid groups from player_groups table')
+    TriggerEvent('qbx_core:server:playergroupscleaned')
```
Manason, line 390: What kind of action would be taken on this event?
```suggestion
    TriggerEvent('qbx_core:server:playerGroupsCleaned')
```

server/storage/players.lua
```diff
@@ -369,6 +372,24 @@ RegisterCommand('convertjobs', function(source)
     TriggerEvent('qbx_core:server:jobsconverted')
 end, true)
 
+---Removes invalid groups from the player_groups table.
+RegisterCommand('cleanplayergroups', function(source)
```
Manason, line 376: Also needs to handle checking for a valid grade as well.

Manason: Also consider adding an opt-in convar to automatically run this on server start. Just need to make sure the jobs/gangs are loaded first.

Manason (changes requested):

server/storage/players.lua
solareon: Was just following the pattern above. probably not too useful

Manason: Should we keep it then? If there's no use case, I would be in favor of removing the event.

server/storage/players.lua
solareon: Fixed. Checks group and group/grade separately and then skips loading if invalid

server/storage/players.lua
solareon: Fixed and convar added.

Manason (approved):
May want to require server.groups to guarantee it is loaded before this file. Right now this is the case as it's loaded at the top of server/main.lua, which is the first file loaded in the fxmanifest. However, that's a bit of a brittle assumption, so it would be good to make this more robust on its own.

mafewtm (approved):

solareon:
> May want to require server.groups to guarantee it is loaded before this file. Right now this is the case as it's loaded at the top of server/main.lua, which is the first file loaded in the fxmanifest. However, that's a bit of a brittle assumption, so it would be good to make this more robust on its own.

That's why I threw the auto clean into that CreateThread at the bottom which should cause it to run at the very end of the resource loading. It's definitely something that could be improved with some guard conditions to make sure the whole table isn't blank or something

PR #528, opened 2024-08-05
feat: ox_inventory native compatibility

solareon:
## Description

Adds support for ox_inventory via qbx specific bridge.

## Checklist


solareon:
Requires https://github.com/solareon/ox_inventory_qbx/tree/qbx-bridge for testing.

server/player.lua
```diff
@@ -830,9 +841,69 @@ function CreatePlayer(playerData, Offline)
...
+    end
+
+    ---@deprecated use ox_inventory exports directly
+    function self.Functions.ClearInventory()
+        return exports.ox_inventory:ClearInventory(self.PlayerData.source)
+    end
+
+    ---@deprecated use ox_inventory exports directly
+    function self.Functions.SetInventory()
+        lib.print.error('Player.Functions.SetInventory is unsupported for ox_inventory. Try ClearInventory, then add the desired items.')
```
Manason, line 904: Should this be error() instead to blowup the stack? Otherwise scripts will assume they've successfully set the inventory and may continue executing in a bad state.

server/player.lua
```diff
@@ -537,6 +541,7 @@ function CheckPlayerData(source, playerData)
         }
     }
     playerData.gangs = gangs or {}
+    playerData.groups = groups or {}
```
Manason, line 544: What's the purpose of this when we already have gangs and jobs tables?

Manason (changes requested):

server/player.lua
solareon: ox_inventory doesn't really differentiate the two, serving them as one combined table simplifies the information transfer to ox_inventory as it also doesn't care about duty status for jobs.

server/player.lua
solareon:
```suggestion
        error('Player.Functions.SetInventory is unsupported for ox_inventory. Try ClearInventory, then add the desired items.')
```

good call. don't want to let stuff get into a bad state.

server/player.lua
Manason: Instead of adding to playerData, which is also apart of the public API for Qbox, maybe this should be a derived field computed within ox_inventory qbx bridge layer?

modules/compat.lua
```diff
@@ -0,0 +1,98 @@
+-- Item conversion borrowed from ox_inventory: https://github.com/overextended/ox_inventory/blob/52d99285eef1dc7df31c084679db3fdf6b1c0150/modules/items/server.lua#L115-L217
+-- Copyright (c) 2023 Overextended
+
+return {
```
Manason, line 4: The modules folder is for publicly available libraries. This should be moved somewhere else. Probably within the bridge.

server/functions.lua
```diff
@@ -436,7 +436,45 @@ exports('ExploitBan', ExploitBan)
...
+---@return PlayerData[]
+local function getPlayersData()
+    local playersData = {}
+    for _, player in pairs(QBX.Players) do
+        playersData[#playersData + 1] = player.PlayerData
+    end
+    return playersData
+end
+
+exports('GetPlayersData', getPlayersData)
```
Manason, line 472: Do we really need this? We already have an export to get the QBX.Players, which is mostly PlayerData already.

server/player.lua
```diff
@@ -744,6 +747,7 @@ function CreatePlayer(playerData, Offline)
             TriggerClientEvent('hud:client:OnMoneyChange', self.PlayerData.source, moneytype, amount, false)
             TriggerClientEvent('QBCore:Client:OnMoneyChange', self.PlayerData.source, moneytype, amount, 'add', reason)
             TriggerEvent('QBCore:Server:OnMoneyChange', self.PlayerData.source, moneytype, amount, 'add', reason)
+            TriggerEvent('qbx_core:server:setMoney', self.PlayerData.source, moneytype, self.PlayerData.money[moneytype])
```
Manason, line 750: Why the new event?

server/storage/players.lua
```diff
@@ -317,6 +317,7 @@ end
 ---@param citizenid string
 ---@return table<string, integer> jobs
 ---@return table<string, integer> gangs
+---@return table<string, integer> playerGroups
```
Manason, line 320: There is no matching return for this annotation.

shared/weapons.lua
```diff
@@ -1,160 +0,0 @@
...
----@field weapontype string
----@field ammotype? string
----@field damagereason string
-
----@deprecated This file is deprecated and will be removed in the future. If you are utilizing QB-Core bridge functionality you will need to populate weapons here for them to be available in QBCore.Shared.Weapons. If not please add your weapons in ox_inventory/data/weapons.lua file. Currently weapons in this file will be converted on next server start.
----@type table<number, Weapon>
-return {
-	-- // WEAPONS
-	-- Melee
-	[`weapon_unarmed`] 				 = {name = 'weapon_unarmed', 		label = 'Fists', 				weapontype = 'Melee',	ammotype = nil, damagereason = 'Melee killed / Whacked / Executed / Beat down / Murdered / Battered'},
```
Manason, line 13: Since ox_inventory format is missing weapontype, ammotype, and damagereason, we need to keep those fields in this file. They can then be combined with the ox_inventory data to recreate the format of QB weapons. Otherwise this will break compatibility with scripts that rely on these fields.

shared/main.lua
```diff
@@ -12,4 +11,18 @@ for _, v in pairs(qbShared.Vehicles) do
 	qbShared.VehicleHashes[v.hash] = v
 end
 
+local weaponConfig = require '@ox_inventory.data.weapons'
+qbShared.Weapons = weaponConfig.Weapons
+qbShared.Components = weaponConfig.Components
+qbShared.AmmoTypes = weaponConfig.Ammo
+
+---@type table<number, Weapon>
+qbShared.WeaponHashes = {}
```
Manason, line 20: This looks like a new table which represents the QB format, but doesn't qbShared.Weapons need to represent the QB format? How are we going to be using these new fields of qbShared?

Manason: Also, changing qbShared at all could be out of scope for this PR if you want to split if off.

Manason (changes requested):

modules/compat.lua
solareon: yeah i just threw it in there as a start. i'll move it

server/functions.lua
solareon: Doesn't this pass all the function tables as well which ox doesn't really need? I guess it's kind of extra

server/storage/players.lua
solareon:
```suggestion
```

shared/main.lua
solareon: the idea was to just make the shared/items.lua be a one way input into ox_inventory format and then we create the qb format for the backwards compatibility.  I hadn't finished the weapons piece which was going to be mostly the stuff from #451

server/player.lua
BerkieBb: To make sure qb users aren't able to use the ox_inventory bridge like linden asked

Manason: Should we call this new event onMoneyChange as well and pass it old amount, new amount, and reason?

solareon: refactored to separate event and calls ox_inventory export directly

Manason (approved):

KostaZx (approved):

PR #531, opened 2024-08-11
feat(types): everything else

mafewtm:
## Description

Finishes adding the rest of the types for `qbx_core`. I intentionally left out the config and module annotations as I figured those should stay where they are. If we think otherwise, I can move those as well.

## Checklist


solareon (approved):
Does this make my editor happy?

Manason (approved):

PR #532, opened 2024-08-12
Fix Player Joining Error

mmoftah:
## Description

Fixes the following error message when a player without a `license2` tries to join the server:

```
[ERROR] @qbx_core/server/storage/players.lua:63: no identifier provided
```

The root cause is: https://github.com/Qbox-project/qbx_core/pull/519. The issue here is that `fetchBan` expects a non-nil license, and since the PR introduces a check for two different licenses as opposed to either license, if a user is missing one of them it will not allow them to join the server and will have a failure message.

Alternate fix:
In `IsPlayerBanned` check if the license is valid before checking. If this approach is preferred let me know and I can quickly make that change.

## Checklist

### DIFF
--- a/server/storage/players.lua
+++ b/server/storage/players.lua
@@ -44,7 +44,7 @@ local function getBanId(request)
     elseif request.ip then
         return 'ip', request.ip
     else
-        error('no identifier provided', 2)
+        return nil, nil
     end
 end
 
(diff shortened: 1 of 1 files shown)


solareon:
This is being worked under #533

PR #533, opened 2024-08-12
fix(server/functions): ban checking when no license2

mafewtm:
## Description

Fixes an issue introduced in #519 where function catches when checking for a player without license2

## Checklist

### DIFF
--- a/server/functions.lua
+++ b/server/functions.lua
@@ -348,7 +348,10 @@ function IsPlayerBanned(source)
         return true, ('You have been banned from the server:\n%s\nYour ban expires in %s/%s/%s %s:%s\n'):format(result.reason, timeTable.day, timeTable.month, timeTable.year, timeTable.hour, timeTable.min)
     else
         CreateThread(function()
-            if license2 then storage.deleteBan({ license = license2 }) end
+            if license2 then
+                storage.deleteBan({ license = license2 })
+            end
+
             storage.deleteBan({ license = license })
         end)
     end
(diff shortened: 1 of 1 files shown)


solareon (approved):
Dealer's choice

server/functions.lua
```diff
@@ -332,29 +332,27 @@ exports('ToggleOptin', ToggleOptin)
...
     if os.time() < result.expire then
         local timeTable = os.date('*t', tonumber(result.expire))
+
         return true, ('You have been banned from the server:\n%s\nYour ban expires in %s/%s/%s %s:%s\n'):format(result.reason, timeTable.day, timeTable.month, timeTable.year, timeTable.hour, timeTable.min)
     else
         CreateThread(function()
-            storage.deleteBan({
-                license = plicense
-            })
+            storage.deleteBan({ license = license2 })
```
mmoftah, line 351: These will cause the same issue as before since `deleteBan` calls `getBanId`

solareon (approved):

server/functions.lua
solareon: Fixed this one with a guard clause

mmoftah:
LGTM, tested locally and it lets me in 🎉

server/functions.lua
```diff
@@ -332,29 +332,30 @@ exports('ToggleOptin', ToggleOptin)
...
     else
         CreateThread(function()
-            storage.deleteBan({
-                license = plicense
-            })
+            if license2 then
+                storage.deleteBan({ license = license2 })
+            end
+
+            storage.deleteBan({ license = license })
```
Manason, line 355: This code implies that license2 is optional, but license is guaranteed. Is that the case in FiveM?

solareon: sadly that's the case. i think it's probably a relic from folks with a REALLY old install of fivem that they just don't get a license2 created at some point

PR #538, opened 2024-08-17
refactor!: move shared to data

solareon:
## Description

Apparently the shared folder is just one that folks skip when they update. This moves those files to the data folder where they can be more immutable. This will obviously break stuff from bad people that do things like `require '@qbx_core.shared.items'` but those should be few and far between.

## Checklist


Manason (approved):

Manason (approved):
These could also be moved to config/shared. Something to consider.

solareon:
> These could also be moved to config/shared. Something to consider.

yeah that's a better place. marking draft for now and i'll fix it.

solareon:
Moved it to config/shared

mafewtm:
I think we should just stick with them being in a data folder like this was originally supposed to be tbh. They aren't configs.

fxmanifest.lua
```diff
@@ -41,13 +41,8 @@ server_scripts {
 
 files {
     'modules/*.lua',
-    'shared/gangs.lua',
-    'shared/items.lua',
-    'shared/jobs.lua',
-    'shared/locations.lua',
+    'data/*.lua',
```
mafewtm, line 44: also this is no longer needed currently

solareon: good catch. i'm not opposed to either solution. just kind of needs to be in it's own place away from actual "code" that needs to execute

PR #540, opened 2024-08-17
fix(server/events): statebag handler for hunger/stress/thirst

solareon:
## Description

Setting the statebags outside of the core never actually updates the metadata values which other resources may depend on. This corrects that with a guard check to prevent infinite recursion.

## Checklist

### DIFF
--- a/server/events.lua
+++ b/server/events.lua
@@ -195,3 +195,29 @@ RegisterNetEvent('QBCore:ToggleDuty', function()
         Notify(src, locale('info.on_duty'))
     end
 end)
+
+---Syncs the player's hunger, thirst, and stress levels with the statebags
+---@param bagName string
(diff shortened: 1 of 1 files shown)


Manason (approved):
Should this PR also include a guard clause to prevent the statebag from being set twice? https://github.com/Qbox-project/qbx_core/blob/main/server/player.lua#L670 Small optimization as it would still terminate after goes around one time.

mafewtm (approved):

Manason (changes requested):
Tested and it results in infinite loop leading to stack overflow.

KostaZx (approved):

mafewtm (approved):
Tested and appears to work as intended

Manason (approved):

PR #545, opened 2024-08-19
feat: currentPropertyId metadata

BerkieBb:
## Description

This add the currentPropertyId metadata for use with qbx_properties. This is to ensure you can use last location inside a property.

## Checklist


solareon (approved):

Manason (commented):
This doesn't seem needed in core as it doesn't have a default value and is modified in the properties PR itself. Would like to close off new additions to the explicit metadata and encourage the way you've done it in the properties PR instead.

BerkieBb:
> This doesn't seem needed in core as it doesn't have a default value and is modified in the properties PR itself. Would like to close off new additions to the explicit metadata and encourage the way you've done it in the properties PR instead.

That's a good one, although the type can then be adjusted to also accept more values that are not standard. Because now it will yell if you check a value that is not in the type

PR #553, opened 2024-08-22
fix(server/queue): get max players via convar for better reliability

D4isDAVID:
## Description

For some reason `GlobalState.MaxPlayers` isn't completely reliable and can sometimes be `nil`, so this gets the convar directly instead.

## Checklist

### DIFF
--- a/server/queue.lua
+++ b/server/queue.lua
@@ -18,7 +18,7 @@ end
 -- Queue code
 
 local config = require 'config.queue'
-local maxPlayers = GlobalState.MaxPlayers
+local maxPlayers = GetConvarInt('sv_maxclients', 48)
 
 -- destructure frequently used config options
 local waitingEmojis = config.waitingEmojis


server/queue.lua
```diff
@@ -18,7 +18,7 @@ end
 -- Queue code
 
 local config = require 'config.queue'
-local maxPlayers = GlobalState.MaxPlayers
+local maxPlayers = GetConvarInt('sv_maxclients', 48)
```
solareon, line 21:
```suggestion
local maxPlayers = GlobalState.MaxPlayers or GetConvarInt('sv_maxclients', 48)
```
Might be slightly more bulletproof?

solareon (approved):

server/queue.lua
D4isDAVID: I don't see a reason for that? `GetConvarInt` will be fully reliable.

mafewtm: if its set and gotten the same way then how is it less reliable?

D4isDAVID: I'm guessing it's a sort of race condition? Not entirely sure why it happens but there's no reason for `GetConvarInt` to fail. I'm guessing using `GlobalState` in this case in general isn't  the most reliable.

mafewtm (approved):

PR #555, opened 2024-08-29
feat: motd

D4isDAVID:
## Description

To improve the onboarding of newbies to Qbox, this PR adds:
- Client MOTD sent in the chat after the player loads in; can be configured.
- Server MOTD printed to the console after the server starts; can be disabled.
- Service messages taken from the `txAdminRecipe`, repo printed to the console after the server starts; can be disabled.

Server MOTD example:
image

See client MOTD and service message examples here: Qbox-project/txAdminRecipe#86

## Checklist

### DIFF
--- a/client/events.lua
+++ b/client/events.lua
@@ -7,6 +7,11 @@ RegisterNetEvent('QBCore:Client:OnPlayerLoaded', function()
         SetCanAttackFriendly(cache.ped, true, false)
         NetworkSetFriendlyFireOption(true)
     end
+
+    local motd = GetConvar('qbx:motd', '')
+    if motd ~= '' then
+        exports.chat:addMessage({ template = motd })
+    end
 end)
 
 ---@param val PlayerData
--- a/server/motd.lua
+++ b/server/motd.lua
@@ -1,0 +1,96 @@
+local acknowledged = GetConvar('qbx:acknowledge', 'false') == 'true'
+local messagesUrl = GetConvar('qbx:serviceMessagesUrl', 'https://raw.githubusercontent.com/Qbox-project/txAdminRecipe/refs/heads/main/service-messages.json')
+local resourceVersion = GetResourceMetadata(cache.resource, 'version', 0)
+
+---@param str string
+---@param prefix string
+---@return boolean
+local function startsWith(str, prefix)
+    return str:sub(1, prefix:len()) == prefix
+end
+
+---Returns whether the given version prefix or exact version matches the resource's version.
+---@return boolean
+local function isResourceVersion(version)
+    if type(version) == 'string' then
+        -- if we don't have a dot at the end, check for an exact version match
+        if version:sub(-1) ~= '.' then
+            return resourceVersion == version
+        end
+
+        -- otherwise treat `version` as a prefix
+        return startsWith(resourceVersion, version)
+    end
+
+    if type(version) == 'table' then
+        for i = 1, #version do
+            local currentVersion = version[i]
+            if isResourceVersion(currentVersion) then
+                return true
+            end
+        end
+
+        return false
+    end
+
+    return not version
+end
+
+---Returns the given content stringified, or concatenated with a newline separator if given a table.
+---@return string
+local function validateContent(content)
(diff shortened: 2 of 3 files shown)


client/events.lua
```diff
@@ -7,6 +7,15 @@ RegisterNetEvent('QBCore:Client:OnPlayerLoaded', function()
         SetCanAttackFriendly(cache.ped, true, false)
         NetworkSetFriendlyFireOption(true)
     end
+
+    local template = GetConvar('qbx:motdTemplate', '')
+    local args = json.decode(GetConvar('qbx:motdArgs', '[]'))
```
Manason, line 12: Is there a way we can make the motd a single string convar? This feels a bit too in the weeds, I assume people could write the style inline if it were a single string?

Manason (changes requested):

client/events.lua
D4isDAVID: That's true, we can make it the template only without the args.

D4isDAVID: Ah and as for the inline style, that doesn't work inside args. The only way to do a single convar is if we make it template-only.

Manason (approved):
Consider adding a line to the readme for this feature (and any other missing features in there while you are editing!)

client/events.lua
Manason: Nice. I think that's a better way to do it.

server/main.lua
```diff
@@ -33,6 +33,19 @@ QBX.Player_Buckets = {}
 QBX.Entity_Buckets = {}
 QBX.UsableItems = {}
 
+if GetConvar('qbx:acknowledge', 'false') ~= 'true' then
+    local welcomeMessage = [[^4
```
solareon, line 37: Not sure if it would be in poor taste but maybe have it pull this from a repo to allow us to send messages to folks about important changes for updates?

solareon (approved):

server/main.lua
Manason: That's not a bad idea.

D4isDAVID: Perhaps our `.github` repo would be appropriate?

solareon: that or maybe txAdminRecipe

server/motd.lua
```diff
@@ -0,0 +1,96 @@
+local acknowledged = GetConvar('qbx:acknowledge', 'false') == 'true'
+local messagesUrl = GetConvar('qbx:serviceMessagesUrl', 'https://raw.githubusercontent.com/Qbox-project/txAdminRecipe/refs/heads/main/service-messages.json')
+local resourceVersion = GetResourceMetadata(cache.resource, 'version', 0)
+
+---@param str string
+---@param prefix string
+---@return boolean
+local function startsWith(str, prefix)
+    return str:sub(1, prefix:len()) == prefix
```
Manason, line 9: Should we handle the case where str is an empty string?

server/motd.lua
```diff
@@ -0,0 +1,96 @@
...
+end
+
+---Returns whether the given version prefix or exact version matches the resource's version.
+---@return boolean
+local function isResourceVersion(version)
```
Manason, line 14: May want to either get rid of the recursion or add a clear param type for version so that it's clear that a table of tables wouldn't be expected.

Manason (approved):

server/motd.lua
D4isDAVID: Currently it will make the message never show up. Should we make it otherwise?

server/motd.lua
D4isDAVID: Technically it *can* work. The problem with adding a clear param type is that we're working with JSON objects, in which case the real type isn't really known. It's also why I handled multiple different types for `version` and `content`.

server/motd.lua
Manason: That's probably fine. As long as it isn't an error condition.

D4isDAVID: From my testing, no inputs should result in errors.

mafewtm (approved):

Issue #559, opened 2024-09-01
Money item is not synchronized


Hunter3001:
### Summary

The item money, is not synced with the actual cash

### Reproduction

- cash received or withdrawn through a script, the item money is not updated.
- command /addmoney cash 1000 the money item is not synced here either

### Expected behavior

I expect the money item to always be kept up to date with the actual cash balance.

### Actual behavior

Hello,
I switched to Qbox when version v1.18.0 was released. Now, with the 1.18.1 update, I hoped that the issue with the Money item would be resolved since it's a significant problem. Unfortunately, the issue still persists—the Money item is not synchronized with the actual cash. For example, when I receive or lose cash, the change is reflected in the HUD, but the Money item in the inventory does not update. I know this issue didn't exist in v1.17.2, but downgrading doesn't make sense for me as several scripts now require v1.18.0. I just wanted to bring this bug to your attention again.

I appreciate your hard work and dedication—keep up the great work!

Best regards

### Current Version

v1.18.1

### Custom Resources

Some, but the problem also exists without custom scripts

solareon:
I trust you are using the latest ox_inventory release as well? Can you please provide the code for `addmoney` as I don't believe this is a baked in command. I just did the following test code on latest core and latest ox_inventory and it worked fine.

```lua
lib.addCommand('addmoney', {
    help = 'add money to your account',
    params = {
        {name = 'type', help = 'the type of money to add'},
        {name = 'amount', help = 'the amount of money to add'}
    },
    restricted = 'group.admin'
}, function(source, args)
    local player = exports.qbx_core:GetPlayer(source)
    player.Functions.AddMoney(args.type, args.amount)
end)

lib.addCommand('addmoneyqb', {
    help = 'add money to your account',
    params = {
        {name = 'type', help = 'the type of money to add'},
        {name = 'amount', help = 'the amount of money to add'}
    },
    restricted = 'group.admin'
}, function(source, args)
    local QBCore = exports['qb-core']:GetCoreObject()
    local player = QBCore.Functions.GetPlayer(source)
    player.Functions.AddMoney(args.type, args.amount)
end)```

@Hunter3001

solareon:
For additional test coverage I did these as well

```lua
lib.addCommand('setmoney', {
    help = 'set money on your account',
    params = {
        {name = 'type', help = 'the type of money to set'},
        {name = 'amount', help = 'the amount of money to set'}
    },
    restricted = 'group.admin'
}, function(source, args)
[...]

PR #561, opened 2024-09-01
feat(bridge/qb): add warnings and disabling of overridden functions

solareon:
QBCore offered a method to modify player object functions at runtime which can lead to interesting side effects depending on implementation. This adds warnings for when this occurs and adds options to block the behavior as well as disabling the warning in either case.

## Description

In cases where resources override existing functions such as https://github.com/Qbox-project/qbx_core/issues/559 we should throw a warning to advise the user that unintended behavior may occur. We can also provide a convar to selectively disable this behavior as well as the warning.

## Checklist

### DIFF
--- a/bridge/qb/server/functions.lua
+++ b/bridge/qb/server/functions.lua
@@ -2,6 +2,9 @@ require 'server.functions'
 require 'bridge.qb.server.player'
 local functions = {}
 
+local allowMethodOverrides = GetConvarInt('qbx:allowMethodOverrides', 1) == 1
+local disableMethodOverrideWarning = GetConvarInt('qbx:disableMethodOverrideWarning', 0) == 1
+
 local createQbExport = require 'bridge.qb.shared.export-function'
 
 ---@deprecated use the GetEntityCoords and GetEntityHeading natives directly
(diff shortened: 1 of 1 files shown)


bridge/qb/server/functions.lua
```diff
@@ -2,6 +2,9 @@ require 'server.functions'
 require 'bridge.qb.server.player'
 local functions = {}
 
+local allowMethodOverrides = GetConvarInt('qbx:allowmethodoverrides', 1) == 1
+local disableMethodOverrideWarning = GetConvarInt('qbx:disableoverridewarning', 0) == 1
```
mafewtm, line 6: these should be `true` or `false`

solareon: https://github.com/overextended/ox_inventory/blob/cf88913cfffe3122a6655ed4af04b18cf159ea11/init.lua#L79-L80

I don't think it works the other way with the `GetConvar()` as it would return a string of `'true'` vice `true`

mafewtm: im saying change it to GetConvar and use boolean values like how we do it [here](https://github.com/Qbox-project/qbx_core/blob/main/server/player.lua#L7)

Manason (approved):

mafewtm (approved):

PR #562, opened 2024-09-07
feat: diagnostic comments

mafewtm:
## Description

- Adds diagnostic comments to make LLS happy
- Adds `NotificationPosition` and `NotificationType` alias to types.lua to satisfy diagnostics

Once merged, https://github.com/Qbox-project/github-docs/pull/17 will satisfy the last remaining `undefined-global` diagnostic warnings.

If any of these should actually be fixed instead of diagnostically dismissed, feel free to comment or add to this PR.

## Checklist

### DIFF
--- a/bridge/qb/server/functions.lua
+++ b/bridge/qb/server/functions.lua
@@ -37,8 +37,7 @@ function functions.SpawnVehicle(source, model, coords, warp)
     local veh = CreateVehicle(model, coords.x, coords.y, coords.z, heading, true, true)
     while not DoesEntityExist(veh) do Wait(0) end
     if warp then
-        ---@diagnostic disable-next-line: missing-parameter
-        while GetVehiclePedIsIn(ped) ~= veh do
+        while GetVehiclePedIsIn(ped, false) ~= veh do
             Wait(0)
             TaskWarpPedIntoVehicle(ped, veh, -1)
         end
--- a/bridge/qb/server/functions.lua
+++ b/bridge/qb/server/functions.lua
@@ -37,6 +37,7 @@ function functions.SpawnVehicle(source, model, coords, warp)
     local veh = CreateVehicle(model, coords.x, coords.y, coords.z, heading, true, true)
     while not DoesEntityExist(veh) do Wait(0) end
     if warp then
+        ---@diagnostic disable-next-line: missing-parameter
         while GetVehiclePedIsIn(ped) ~= veh do
             Wait(0)
             TaskWarpPedIntoVehicle(ped, veh, -1)
@@ -436,6 +437,7 @@ function functions.AddPlayerMethod(ids, methodName, handler)
         end
     elseif idType == 'table' and table.type(ids) == 'array' then
         for i = 1, #ids do
+            ---@diagnostic disable-next-line: deprecated
             functions.AddPlayerMethod(ids[i], methodName, handler)
         end
     end
@@ -457,15 +459,18 @@ function functions.AddPlayerField(ids, fieldName, data)
     if idType == 'number' then
         if ids == -1 then
             for _, v in pairs(QBX.Players) do
+                ---@diagnostic disable-next-line: undefined-field
                 v.Functions.AddField(fieldName, data)
             end
         else
             if not QBX.Players[ids] then return end
 
+            ---@diagnostic disable-next-line: undefined-field
             QBX.Players[ids].Functions.AddField(fieldName, data)
         end
     elseif idType == 'table' and table.type(ids) == 'array' then
         for i = 1, #ids do
+            ---@diagnostic disable-next-line: deprecated
             functions.AddPlayerField(ids[i], fieldName, data)
         end
     end
@@ -520,7 +525,7 @@ function functions.GetSource(identifier)
 end
 
 ---@param source Source|string source or identifier of the player
----@return Player
+---@return Player?
 function functions.GetPlayer(source)
     return AddDeprecatedFunctions(exports.qbx_core:GetPlayer(source))
 end
--- a/bridge/qb/server/main.lua
+++ b/bridge/qb/server/main.lua
@@ -6,6 +6,7 @@ require 'bridge.qb.server.events'
 local convertItems = require 'bridge.qb.shared.compat'.convertItems
 convertItems(require '@ox_inventory.data.items', require 'shared.items')
 
+---@diagnostic disable-next-line: lowercase-global
 qbCoreCompat = {}
 
 qbCoreCompat.Config = lib.table.merge(require 'config.server', require 'config.shared')
@@ -19,6 +20,8 @@ qbCoreCompat.Entity_Buckets = QBX.Entity_Buckets
 qbCoreCompat.UsableItems = QBX.UsableItems
 qbCoreCompat.Functions = require 'bridge.qb.server.functions'
 qbCoreCompat.Commands = require 'bridge.qb.server.commands'
+
+---@diagnostic disable: deprecated
 
 ---@deprecated Call lib.print.debug() instead
 qbCoreCompat.Debug = lib.print.debug
--- a/bridge/qb/server/player.lua
+++ b/bridge/qb/server/player.lua
@@ -86,9 +86,8 @@ end
 ---Will cause major issues!
 ---@param playerData PlayerData
 ---@param Offline boolean
----@return Player player
+---@return Player?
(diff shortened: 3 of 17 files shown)


bridge/qb/shared/main.lua
```diff
@@ -85,12 +87,16 @@ qbShared.CommaValue = lib.math.groupdigits
 qbShared.RandomStr = function(length)
     if length <= 0 then return '' end
     local pattern = math.random(2) == 1 and 'a' or 'A'
+
+    ---@diagnostic disable-next-line: deprecated
     return qbShared.RandomStr(length - 1) .. lib.string.random(pattern)
```
Manason, line 92: Instead of recursively calling this function, this should construct a string of '.' of length length, and only call lib.string.random() to avoid the deprecated call.

bridge/qb/server/functions.lua
```diff
@@ -37,6 +37,7 @@ function functions.SpawnVehicle(source, model, coords, warp)
     local veh = CreateVehicle(model, coords.x, coords.y, coords.z, heading, true, true)
     while not DoesEntityExist(veh) do Wait(0) end
     if warp then
+        ---@diagnostic disable-next-line: missing-parameter
         while GetVehiclePedIsIn(ped) ~= veh do
```
Manason, line 41: Can just add the missing parameter?

bridge/qb/server/main.lua
```diff
@@ -20,6 +21,8 @@ qbCoreCompat.UsableItems = QBX.UsableItems
 qbCoreCompat.Functions = require 'bridge.qb.server.functions'
 qbCoreCompat.Commands = require 'bridge.qb.server.commands'
 
+---@diagnostic disable: deprecated
+
 ---@deprecated Call lib.print.debug() instead
```
Manason, line 26: Curious what is triggering warnings about this line?

bridge/qb/server/player.lua
```diff
@@ -88,7 +88,8 @@ end
 ---@param Offline boolean
 ---@return Player player
 function playerObj.CreatePlayer(playerData, Offline)
-    AddDeprecatedFunctions(exports.qbx_core:CreatePlayer(playerData, Offline))
+    ---@diagnostic disable-next-line: return-type-mismatch
```
Manason, line 91: The return type shouldn't mismatch. I think the return type Just needs to be changed to Player?

modules/playerdata.lua
```diff
@@ -4,6 +4,7 @@ QBX = {} -- luacheck: ignore
 QBX.PlayerData = exports.qbx_core:GetPlayerData() or {}
 
 RegisterNetEvent('QBCore:Client:OnPlayerUnload', function()
+    ---@diagnostic disable-next-line: missing-fields
     QBX.PlayerData = {}
```
Manason, line 8: Maybe we should change the definition of QBX.PlayerData to include an empty table as an allowed value for the type?

Manason (approved):

bridge/qb/server/functions.lua
mafewtm: Seeing as this is in the bridge, I opted to add the diagnostic instead of adding the parameter as qb-core has it just like this. If we think it'd be better to add the `lastVehicle` boolean instead, we can.

bridge/qb/server/main.lua
mafewtm: All the `qbCoreCompat.Functions...` and `qbCoreCompat` callbacks in the code below. I just figured it'd be better to toss the diagnostic for it all instead of going line by line for each of them.

modules/playerdata.lua
mafewtm: Where is it even corely defined?

bridge/qb/server/functions.lua
D4isDAVID: I assume a lack of a parameter here is the same as `false`? Seems to also be the intention

bridge/qb/shared/main.lua
```diff
@@ -85,13 +87,15 @@ qbShared.CommaValue = lib.math.groupdigits
 qbShared.RandomStr = function(length)
     if length <= 0 then return '' end
     local pattern = math.random(2) == 1 and 'a' or 'A'
-    return qbShared.RandomStr(length - 1) .. lib.string.random(pattern)
+
+    return lib.string.random(pattern, length)
```
D4isDAVID, line 91: Careful! The docs on [`lib.string.random`](https://overextended.dev/ox_lib/Modules/String/Shared#stringrandom) state that the `length` argument is only for padding it or omitting characters! It does not affect the pattern in any other way.
That is not the intention of `RandomStr`, so you should revert to the previous solution or create the pattern with a loop.

bridge/qb/shared/main.lua
```diff
@@ -85,13 +87,15 @@ qbShared.CommaValue = lib.math.groupdigits
...
 end
 
 ---@deprecated use lib.string.random from ox_lib
 qbShared.RandomInt = function(length)
     if length <= 0 then return '' end
-    return qbShared.RandomInt(length - 1) .. lib.string.random('1')
+
+    return lib.string.random('.', length)
```
D4isDAVID, line 98: Ditto, and you should be using `1` for the pattern, since `.` is both random letters and numbers, while `1` is only random numbers.

client/functions.lua
```diff
@@ -55,6 +55,7 @@ exports('HasPrimaryGroup', HasPrimaryGroup)
 ---@param filter string | string[] | table<string, number>
 ---@return boolean
 function HasGroup(filter)
+    ---@diagnostic disable-next-line: missing-parameter
```
D4isDAVID, line 58: Can we make the 3rd parameter for `HasPlayerGotGroup` optional instead?

modules/utils.lua
```diff
@@ -147,6 +149,7 @@ if isServer then
         model = type(model) == 'string' and joaat(model) or (model --[[@as integer]])
         local ped = GetPlayerPed(source)
 
+        ---@diagnostic disable-next-line: redundant-return-value
```
D4isDAVID, line 152: It might be better to *actually* avoid returning more than we claim to. It'd probably be annoying to debug in certain use cases. So just save the first value in a variable and return it, ignoring the second one.

server/functions.lua
```diff
@@ -408,6 +416,7 @@ exports('GetCoreVersion', GetCoreVersion)
 local function ExploitBan(playerId, origin)
     local name = GetPlayerName(playerId)
     local success, errorResult = pcall(
+        ---@diagnostic disable-next-line: param-type-mismatch
```
D4isDAVID, line 419: What is even going on here? Is this code even valid? `pcall` is supposed to execute a function given to it.

server/functions.lua
```diff
@@ -446,6 +455,8 @@ exports('HasPrimaryGroup', HasPrimaryGroup)
 ---@return boolean
 function HasGroup(source, filter)
     local playerData = QBX.Players[source].PlayerData
+
+    ---@diagnostic disable-next-line: missing-parameter
```
D4isDAVID, line 459: Ditto on making the 3rd param optional for `HasPlayerGotGroup`.

D4isDAVID (changes requested):

server/functions.lua
mafewtm: Absolutely no idea

modules/utils.lua
```diff
@@ -147,12 +149,14 @@ if isServer then
         model = type(model) == 'string' and joaat(model) or (model --[[@as integer]])
         local ped = GetPlayerPed(source)
 
-        return qbx.spawnVehicle({
+        local netId, _ = qbx.spawnVehicle({
```
D4isDAVID, line 152:
```suggestion
        local netId = qbx.spawnVehicle({
```
Can this be omitted? Or is that another diagnostic warning?

mafewtm: It can be omitted. I Just do that out of habit to show that it does in fact return two values but we only use one.

D4isDAVID: Makes sense.

server/functions.lua
D4isDAVID: Just remove the `pcall` part, which will also let us remove this diagnostic comment. The `success` and `errorResult` come from the `insertBan` function directly.

solareon (approved):

PR #564, opened 2024-09-10
feat(server/main): check that one sync infinity is running on server start

Manason:


D4isDAVID:
I'm actually not sure whether that convar is even usable, and the `+set onesync on` argument should probably be preferred when starting FXServer.

mafewtm (approved):

D4isDAVID:
Didn't get the chance to answer, but looks alright.

PR #565, opened 2024-09-10
feat(modules/lib): qbx.array.contains to check if an array contains a given value

Manason:
Uses iterative for loop for better performance compared to ox_lib's table.contains

### DIFF
--- a/modules/lib.lua
+++ b/modules/lib.lua
@@ -147,6 +147,20 @@ function qbx.table.mapBySubfield(tble, subfield)
     return map
 end
 
+---Returns true if the given array contains the given value.
+---@param arr any[]
+---@param val any
+---@return boolean
+function qbx.array.contains(arr, val)
+    for i = 1, #arr do
+        if arr[i] == val then
+            return true
+        end
+    end
+
+    return false
+end
+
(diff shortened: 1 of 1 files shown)


D4isDAVID (changes requested):
I don't believe we define `qbx.array = {}` anywhere yet. Either add that definition at the top or change the function to be something like `qbx.table.arrayContains`.

Manason:
> I don't believe we define `qbx.array = {}` anywhere yet. Either add that definition at the top or change the function to be something like `qbx.table.arrayContains`.

Done.

modules/lib.lua
```diff
@@ -147,6 +148,20 @@ function qbx.table.mapBySubfield(tble, subfield)
     return map
 end
 
+---Returns true if the given array contains the given value.
+---@param arr any[]
+---@param val any
```
D4isDAVID, line 153: Maybe consider adding a generic here instead of `any`.

D4isDAVID (approved):

modules/lib.lua
Manason: Good call, done.

mafewtm (approved):

PR #568, opened 2024-09-14
feat: persistent player vehicles

### LINKED ISSUE #567: Persistent Vehicles
### The problem

There is a bug in FiveM that makes it so server spawned vehicles can be deleted. This can happen even when a player is driving it.

### Ideal solution

When a vehicle is deleted, it is respawned at the position it was deleted, with the same vehicle properties.

Using the server side entityRemoved event, we can intercept the entity deletion to cache some data about the vehicle, such as the vehicleId, plate, and coords. Then, we can lookup the vehicle in the database to determine whether it is an owned vehicle and respawn it. This could result in stale vehicle properties however, as the database may not have been recently updated.

Therefore, in addition to the above solution, we should also save vehicle properties to the database on an on-demand basis. That is, whenever it changes. Only a client can read the vehicle properties, and would need to send this to the server when it changed. Therefore I propose that when a client is occupying the driver seat of a vehicle, regardless of whether it is the entity owner or not it should do the following:

1. cache the vehicle properties
2. every few seconds, check if the vehicle properties have changed
3. send the diff (if there is one) to the server

The server will then

1. receive the diff from the client
2. read the vehicle from the database
3. overwrite diffed fields
4. save the vehicle properties back

If this proves to be too much of a performance cost, the server could save to a buffer, and flush the buffer on a less regular cadence, but this should only be done if there proves to be an issue, not before as it adds its own complications, drawbacks, and edge cases.

Exploits in this solution aren't entirely preventable, as the client is the source of truth for the vehicle property data. To prevent exploits to the extent possible, the following can be done:

- Only fields changed outside of other events need be accepted by the server. Changing the owner of a vehicle, or its mods for example should result in a separate event on its own rather than be changed through a regular properties detection solution. The vehicles the server could accept would be limited to damage and fuel fields.
- if there is a non RPC server native to get the data, the server will use it rather than trust the value sent by the client
- anomalous diffs could be detected. i.e. if a vehicle is repaired, we might expect the event to be sent by the script that did the repairing, and so a regular client diff would point to lua injection. This isn't recommended however, as it would assume that all scripts which repair vehicles would be sending property update events to the server.

### Alternative solutions

_No response_

### Additional context

_No response_

Manason:
fixes #567 

Implements design as described in #567
- Tested
- persistence is disabled by default, requires convar to opt-in.
- Changed places where we delete a vehicle in core to use the new core export.
- We'll need to change other Qbox resources to use the DeleteVehicle export to delete vehicles. This export allows the vehicle to be deleted without being respawned by persistence and works even if persistence is disabled via convar.

Adds new exports
- EnablePersistence
- DisablePersistence
- DeleteVehicle

### DIFF
--- a/client/vehicle-persistence.lua
+++ b/client/vehicle-persistence.lua
@@ -42,6 +42,7 @@ local function calculateDiff(tbl1, tbl2)
 end
 
 local function sendPropsDiff()
+    if not Entity(vehicle).state.persisted then return end
     local newProps = lib.getVehicleProperties(vehicle)
     if not cachedProps then
         cachedProps = newProps
--- a/client/vehicle-persistence.lua
+++ b/client/vehicle-persistence.lua
@@ -1,0 +1,73 @@
+if GetConvar('qbx:enable_vehicle_persistence', 'true') == 'false' then return end
+
+local cachedProps
+local netId
+local vehicle
+local seat
+
+local watchedKeys = {
+    'bodyHealth',
+    'engineHealth',
+    'tankHealth',
+    'fuelLevel',
+    'oilLevel',
+    'dirtLevel',
+    'windows',
+    'doors',
+    'tyres',
+}
+
+---Calculates the difference in values of two tables for the watched keys.
+---If the second table does not have a value that the first table has, it will be marked 'deleted'.
+---@param tbl1 table
+---@param tbl2 table
+---@return table diff
+---@return boolean hasChanged if diff table is not empty
+local function calculateDiff(tbl1, tbl2)
+    local diff = {}
+    local hasChanged = false
+
+    for i = 1, #watchedKeys do
+        local key = watchedKeys[i]
+        local val1 = tbl1[key]
+        local val2 = tbl2[key]
+
+        if val1 ~= val2 then
+            diff[key] = val2 == nil and 'deleted' or val2
+            hasChanged = true
+        end
+    end
+
+    return diff, hasChanged
+end
+
+local function sendPropsDiff()
+    local newProps = lib.getVehicleProperties(vehicle)
+    if not cachedProps then
+        cachedProps = newProps
+        return
+    end
+    local diff, hasChanged = calculateDiff(cachedProps, newProps)
+    cachedProps = newProps
+    if not hasChanged then return end
+    TriggerServerEvent('qbx_core:server:vehiclePropsChanged', netId, diff)
+end
+
+lib.onCache('seat', function(newSeat)
+    if newSeat == -1 then
+        seat = -1
+        vehicle = cache.vehicle
+        netId = NetworkGetNetworkIdFromEntity(vehicle)
+        CreateThread(function()
+            while seat == -1 do
+                sendPropsDiff()
+                Wait(10000)
+            end
+        end)
+    else
+        seat = nil
+        sendPropsDiff()
+        vehicle = nil
+        netId = nil
+    end
+end)
--- a/client/vehicle-persistence.lua
+++ b/client/vehicle-persistence.lua
@@ -64,7 +64,7 @@ lib.onCache('seat', function(newSeat)
                 Wait(10000)
             end
         end)
-    else
+    elseif seat == -1 then
         seat = nil
         sendPropsDiff()
         vehicle = nil
--- a/client/vehicle-persistence.lua
+++ b/client/vehicle-persistence.lua
@@ -1,4 +1,4 @@
-if GetConvar('qbx:enable_vehicle_persistence', 'true') == 'false' then return end
+if GetConvar('qbx:enableVehiclePersistence', 'false') == 'false' then return end
 
 local cachedProps
 local netId
--- a/fxmanifest.lua
+++ b/fxmanifest.lua
@@ -65,7 +65,6 @@ dependencies {
     '/onesync',
     'ox_lib',
     'oxmysql',
-    'qbx_vehicles',
 }
 
 provide 'qb-core'
--- a/fxmanifest.lua
+++ b/fxmanifest.lua
@@ -23,6 +23,7 @@ client_scripts {
     'client/events.lua',
     'client/character.lua',
     'client/discord.lua',
+    'client/vehicle-persistence.lua',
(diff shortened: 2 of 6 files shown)


mafewtm:
I don't think qbx_core should be dependent on qbx_vehicles

Manason:
> I don't think qbx_core should be dependent on qbx_vehicles

qbx_vehicles is an API resource so I think it's fine to depend on it. The alternative is to make qbx_vehicles have a client component instead of server o Lu. Do you think that would be better?

fxmanifest.lua
```diff
@@ -63,6 +65,7 @@ dependencies {
     '/onesync',
     'ox_lib',
     'oxmysql',
+    'qbx_vehicles',
```
solareon, line 68: Does calling this out specifically as a dependency in the manifest affect the start order?

client/vehicle-persistence.lua
```diff
@@ -0,0 +1,73 @@
...
+        netId = NetworkGetNetworkIdFromEntity(vehicle)
+        CreateThread(function()
+            while seat == -1 do
+                sendPropsDiff()
+                Wait(10000)
+            end
+        end)
+    else
+        seat = nil
+        sendPropsDiff()
```
solareon, line 69: Is this needed here? Would this create a race condition with multiple people getting into the same vehicle?

solareon (changes requested):

fxmanifest.lua
Manason: Yes. I'm taking the stance that we need to start qbx_vehicles before core.

Manason: On second thought, removing from manifest as restarting qbx_vehicles shouldn't stop core.

client/vehicle-persistence.lua
Manason: Corrected the condition. This is meant to break the loop only for the driver. Nice catch.

mafewtm (approved):

client/vehicle-persistence.lua
```diff
@@ -0,0 +1,73 @@
...
+                Wait(10000)
+            end
+        end)
+    elseif seat == -1 then
+        seat = nil
+        sendPropsDiff()
+        vehicle = nil
+        netId = nil
+    end
+end)
```
solareon, line 73: Perhaps this could also check if there is a statebag associated with that entity to have it persist before sending any cross network traffic related to vehicles that will be discarded?

Manason: Done. Good call on the state bag instead of the server table for tracking the persisted list.

client/vehicle-persistence.lua
solareon:
```suggestion
```
I think you meant to remove this.

client/vehicle-persistence.lua
```diff
@@ -0,0 +1,74 @@
...
+        return
+    end
+    local diff, hasChanged = calculateDiff(cachedProps, newProps)
+    cachedProps = newProps
+    if not hasChanged then return end
+    TriggerServerEvent('qbx_core:server:vehiclePropsChanged', netId, diff)
+end
+
+lib.onCache('seat', function(newSeat)
+    if newSeat == -1 then
```
solareon, line 58:
```suggestion
lib.onCache('seat', function(newSeat)
	if not Entity(cache.vehicle).state.persisted then return end
    if newSeat == -1 then
```
We shouldn't create a thread and network traffic for vehicles we are not interested in saving.

solareon (changes requested):

client/vehicle-persistence.lua
solareon: n/m

modules/lib.lua
```diff
@@ -419,6 +419,7 @@ else
     end
 
     ---Deletes the specified vehicle and returns whether it was successful.
+    ---@deprecated use exports.qbx_core:DeleteVehicle(vehicle) instead
```
D4isDAVID, line 422: `qbx.deleteVehicle` is client-side. Should we still deprecate it in favor of a server-side export? Not all cars will necessarily be persistent.

server/vehicle-persistence.lua
```diff
@@ -0,0 +1,108 @@
...
+
+        props.tyres = damage
+    end
+
+    exports.qbx_vehicles:SaveVehicle(vehicle, {
+        props = props,
+    })
+end)
+
+AddEventHandler('qbx_core:server:vehicleSpawned', function(entity)
```
D4isDAVID, line 83: What is this event for? Do we want it just to set `EnablePersistence` or is this for other resources to use too? If it's the former then we should just use `EnablePersistence`.

D4isDAVID (commented):
Should we maybe also warp passengers back to the new vehicle?

server/vehicle-persistence.lua
Manason: I was thinking other resources, as one enabling persistence would want it enabled by default on all player vehicles.

modules/lib.lua
Manason: reverted the deprecation.

Manason:
> Should we maybe also warp passengers back to the new vehicle?

Done. Still need to test this however.

server/vehicle-persistence.lua
D4isDAVID: I'm failing to see a use case for this. Specifically for adding another handler to this event, since instead of triggering it you can just use the export.

Manason: This is an internal event for core. Since lib.spawnVehicle is a library, when a vehicle is spawned an event is needed to tell core to persist the vehicle. Otherwise, callers would have to manually call the export after spawning a vehicle (is this what you are suggesting we do instead?)

D4isDAVID: I'm saying wrapping it in an event feels unnecessary. Why not use the export directly in the lib?

Manason: Switched to using the exports and got rid of the event. Make it so exports are always available, instead of only available if persistence is enabled.

solareon (approved):

PR #576, opened 2024-09-20
feat(server/functions): added SearchPlayers export

PhoenixVeil:
## Description

Added the SearchPlayers export, which enables developers to access PlayerData for players in the database without needing their specific identifier and gives more flexibility to developers for advanced systems. The function has filters, so it doesn't dump a bunch of SQL data that is not needed.

## Checklist

### DIFF
--- a/server/functions.lua
+++ b/server/functions.lua
@@ -478,13 +478,6 @@ end
 
 exports('GetPlayersData', getPlayersData)
 
-local function convertPosition(position)
-    local defaultSpawn = require 'config.shared'.defaultSpawn
-    local pos = json.decode(position)
-    local actualPos = (not pos.x or not pos.y or not pos.z) and defaultSpawn or pos
-    return vec4(actualPos.x, actualPos.y, actualPos.z, actualPos.w or defaultSpawn.w)
-end
-
 ---@param filters table<string, any>
 local function handleFilters(filters)
     if not (filters) then return '', {} end
@@ -514,20 +507,17 @@ end
 local function searchPlayerEntities(filters)
     ---@type PlayerEntity[]
     local result = {}
-    local query = "SELECT citizenid, charinfo, money, job, gang, position, metadata, UNIX_TIMESTAMP(last_logged_out) AS lastLoggedOutUnix FROM players"
+    local query = "SELECT citizenid FROM players"
     local where, holders = handleFilters(filters)
     lib.print.debug(query .. where)
     ---@type PlayerEntityDatabase[]
     local response = MySQL.query.await(query .. where, holders)
     for i=1, #response do
-        result[i] = response[i]
-        result[i].charinfo = json.decode(response[i].charinfo)
-        result[i].money = json.decode(response[i].money)
-        result[i].job = response[i].job and json.decode(response[i].job)
-        result[i].gang = response[i].gang and json.decode(response[i].gang)
-        result[i].position = convertPosition(response[i].position)
-        result[i].metadata = json.decode(response[i].metadata)
-        result[i].lastLoggedOut = response[i].lastLoggedOutUnix
+        local citizenid = response[i].citizenid
+        local player = GetOfflinePlayer(citizenid)
+        if player then
+            table.insert(result, player)
+        end
     end
     return result
 end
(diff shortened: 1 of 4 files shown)


Manason (changes requested):
This should probably just return citizen ids matching the search criteria, then do a secondary internal function call to get the player or offline player by citizen id. That way we still benefit from the existing code which constructs player objects properly and adds safety checks and such.

Alternatively, we can still do playerdata only with filtering as this PR does, but it should still go through normal playerdata construction functionality that is shared.

PhoenixVeil:
> This should probably just return citizen ids matching the search criteria, then do a secondary internal function call to get the player or offline player by citizen id. That way we still benefit from the existing code which constructs player objects properly and adds safety checks and such.
> 
> Alternatively, we can still do playerdata only with filtering as this PR does, but it should still go through normal playerdata construction functionality that is shared.

I have updated the function to just fetch the citizenids for the search criteria and then use the Internal Function `GetOfflinePlayer` to handle the rest

server/functions.lua
```diff
@@ -478,6 +478,57 @@ end
...
 
+---@param filters table<string, any>
+local function handleFilters(filters)
+    if not (filters) then return '', {} end
+    local holders = {}
+    local clauses = {}
+    if filters.citizenid then
+        clauses[#clauses + 1] = 'citizenid = ?'
+        holders[#holders + 1] = filters.citizenid
+    end
```
Manason, line 489: I don't think we need this as you don't need to search if you already have a specific citizenid

server/functions.lua
```diff
@@ -478,6 +478,57 @@ end
...
+    local response = MySQL.query.await(query .. where, holders)
+    for i = 1, #response do
+        local citizenid = response[i].citizenid
+        local player = GetPlayerByCitizenId(citizenid)
+        if player then
+            table.insert(result, player)
+        else
+            local offlinePlayer = GetOfflinePlayer(citizenid)
+            if offlinePlayer then
+                table.insert(result, offlinePlayer)
```
Manason, line 523: https://manason.github.io/effective-fivem-lua/tables/#avoid-tableinsert

Manason (changes requested):
Most of this should be moved into storage.lua.

PhoenixVeil:
> Most of this should be moved into storage.lua.

I have moved most of it into the storage/players.lua as requested

server/storage/players.lua
```diff
@@ -128,6 +128,69 @@ local function fetchPlayerEntity(citizenId)
...
+        local player = GetPlayerByCitizenId(citizenid)
+        if player then
+            result[#result+1] = player
+        else
+            local offlinePlayer = GetOfflinePlayer(citizenid)
+            if offlinePlayer then
+                result[#result+1] = offlinePlayer
+            end
+        end
+    end
```
Manason, line 190: This part should be moved back to the exported function. Everything above should remain storage.

server/storage/players.lua
```diff
@@ -128,6 +128,69 @@ local function fetchPlayerEntity(citizenId)
...
+    if filters.job then
+        clauses[#clauses + 1] = 'JSON_EXTRACT(job, "$.name") = ?'
+        holders[#holders + 1] = filters.job
+    end
+    if filters.gang then
+        clauses[#clauses + 1] = 'JSON_EXTRACT(gang, "$.name") = ?'
+        holders[#holders + 1] = filters.gang
+    end
+    if filters.metadata then
+        local typeof, value = filters.metadata.typeof, filters.metadata.value
```
Manason, line 149: Could the type be derived from the value instead of passed?

Manason (changes requested):

server/storage/players.lua
PhoenixVeil: Could you elaborate on this one?
As typeof is the metadata key to fetch like "ishandcuffed" "inlaststand" etc
value is just the value the user is looking for like "true" "false" "1" etc

Manason: Oh. In that case wouldn't it make more sense to use a key value table for metadata rather than two named keys (typeof, and value)?

PhoenixVeil: Done as requested

Manason (approved):

mafewtm (approved):

PR #582, opened 2024-09-25
feat: entity session id

Manason:
We need a stable vehicle identifier that works for both NPC and player owned vehicles. vehicleid already represents player owned vehicles, so overloading it could cause issues with scripts relying on this behavior. netId would not be stable for persisted vehicles, which have their entities deleted and spawned.

Therefore this PR is introducing a sessionId, which will be incremented starting by 1 to guarantee uniqueness. This will be set via a new export CreateSessionId on the vehicle's state bag, which acquires a lock to guarantee that no two vehicles are assigned the same id. This will be persisted during vehicle persistence.

Also sneaking in a fix to persist vehicleid during vehicle persistence as I realize we never did that before.

While this is being introduced for vehicles, it is being introduced as a generic feature that could be used for any entity. This may prove useful for other applications which want a unique id to reference when those entities may be deleted and re-created.

### DIFF
--- a/server/main.lua
+++ b/server/main.lua
@@ -41,23 +41,16 @@ QBX.UsableItems = {}
 local vehicleClasses
 local vehicleClassesPromise
 
-local currentSessionId = 1
-local sessionIdLock = false
+local currentSessionId = 0
 
 ---Sets a unique sessionId statebag on the entity.
 ---@param entity number
 ---@return integer sessionId
 local function createSessionId(entity)
-    if sessionIdLock then
-        lib.waitFor(function()
-            if not sessionIdLock then return true end
-        end, 'timed out acquiring sessionId lock')
-    end
-    sessionIdLock = true
-    Entity(entity).state:set('sessionId', currentSessionId, true)
     currentSessionId += 1
-    sessionIdLock = false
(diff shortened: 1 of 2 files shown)


server/main.lua
```diff
@@ -41,6 +41,27 @@ QBX.UsableItems = {}
...
 
+local currentSessionId = 1
+local sessionIdLock = false
+
+---Sets a unique sessionId statebag on the entity.
+---@param entity number
+---@return integer sessionId
+local function createSessionId(entity)
+    if sessionIdLock then
+        lib.waitFor(function()
```
D4isDAVID, line 52: Is it wild to assume there may be multiple different instances wanting to create a session ID at the same time? In such a case they will all wait together, resolve together, and possibly receive the same session ID. Perhaps some kind of queue should be made instead? Or handle it with promises?

Manason: Yeah you are correct

Manason: Per offline discussion and testing, we're unable to create a scenario where two sessionIds are duplicate. Thus the conclusion is that this is either impossible in the five runtime, or so unlikely as not be worth handling.

D4isDAVID (approved):

PR #584, opened 2024-09-25
feat(modules): Scaleform class

Mycroft-Studios:
## Description

Introduces a scaleform class, based upon  ox lib classes, to make scaleforms both easier to read and use

## Checklist

### DIFF
--- a/modules/scaleform.lua
+++ b/modules/scaleform.lua
@@ -1,38 +1,49 @@
----@class Scaleform : OxClass
+---@class renderTargetTable
 ---@field name string
+---@field model string|number
+
+---@class detailsTable
+---@field name string
+---@field fullScreen? boolean
+---@field x? number
+---@field y? number
+---@field width? number
+---@field height? number
+---@field renderTarget? renderTargetTable
+
+---@class Scaleform : OxClass
 ---@field scaleform number
 ---@field draw boolean
 ---@field target number
 ---@field targetName string
 ---@field handle number
+---@field fullScreen boolean
 local Scaleform = lib.class('Scaleform')
 
----@param details table | string
+---@param details detailsTable | string
 ---@return nil
 ---@description Create a new scaleform class
 function Scaleform:constructor(details)
-    local detailProvided = type(details) == "table"
-    self.name = detailProvided and details.name or details -- Set the name
+    details = type(details) == "table" and details or {name = details} -- Set the details to a table if it is not already
 
-    local scaleform = lib.requestScaleformMovie(self.name) -- Request the scaleform movie
+    local scaleform = lib.requestScaleformMovie(details.name) -- Request the scaleform movie
 
     if not scaleform then -- If the scaleform is nil
-        return error(('Failed to request scaleform movie - [%s]'):format(self.name)) -- Error the failed scaleform request
+        return error(('Failed to request scaleform movie - [%s]'):format(details.name)) -- Error the failed scaleform request
     end
 
     self.handle = scaleform -- Set the scaleform handle
     self.draw = false -- Set the draw to false
-    self.fullScreen = detailProvided and details.fullScreen or true -- Set default for full screen
 
-    if detailProvided then
-        self.x = details.x
-        self.y = details.y
-        self.width = details.width
-        self.height = details.height
+    -- Set Default Values if not provided
+    self.fullScreen = details.fullScreen ~= nil and details.fullScreen or true
+    self.x = details.x or 0
+    self.y = details.y or 0
+    self.width = details.width or 0
+    self.height = details.height or 0
 
-        if details.renderTarget then
-            self:renderTarget(details.renderTarget.name, details.renderTarget.model)
-        end
+    if details.renderTarget then
+        self:setRenderTarget(details.renderTarget.name, details.renderTarget.model)
     end
 end
 
@@ -55,7 +66,7 @@ local function convertArgs(argsTable)
     end
 end
 
----@param type 'bool' | 'int' | 'string'
+---@param type 'boolean' | 'integer' | 'string'
 ---@return boolean | integer | string
 ---@description Awaits the return value, and converts it to a usable data type
 local function retrieveReturnValue(type)
@@ -67,9 +78,9 @@ local function retrieveReturnValue(type)
         end
     end, "Failed to retrieve return value", 1000)
 
-    if type == "int" then -- If the type is an integer
+    if type == "integer" then -- If the type is an integer
         return GetScaleformMovieMethodReturnValueInt(result) -- Get the return value as an integer
-    elseif type == "bool" then
+    elseif type == "boolean" then
         return GetScaleformMovieMethodReturnValueBool(result) -- Get the return value as a boolean
     else -- If the type is not an integer
         return GetScaleformMovieMethodReturnValueString(result) -- Get the return value as a string
@@ -81,7 +92,7 @@ end
 ---@param returnValue? string
 ---@return any
 ---@description Call a scaleform function, with optional args or return value.
-function Scaleform:method(name, args, returnValue)
+function Scaleform:callMethod(name, args, returnValue)
     if not self.handle then -- If the scaleform handle is nil
         return error('Scaleform handle is nil') -- Error the scaleform handle is nil
     end
@@ -132,7 +143,7 @@ end
 ---@param model string|number
 ---@return nil
 ---@description Create a render target for the scaleform - optional , only if you want to render the scaleform in 3D
-function Scaleform:renderTarget(name, model)
+function Scaleform:setRenderTarget(name, model)
 
     -- ensures theres no Targets still active, since this could cause a memory leak
     -- if the render targets are not released.
--- a/modules/scaleform.lua
+++ b/modules/scaleform.lua
@@ -61,7 +61,6 @@ end
 local function retrieveReturnValue(type)
     local result = EndScaleformMovieMethodReturnValue() -- End the scaleform movie method with a return value
 
-    local timeout = 0
     lib.waitFor(function()
         if IsScaleformMovieMethodReturnValueReady(result) then
             return true
--- a/modules/scaleform.lua
+++ b/modules/scaleform.lua
@@ -14,7 +14,7 @@ function Scaleform:constructor(name)
     self.name = name -- Set the name
 
     local scaleform = lib.requestScaleformMovie(name) -- Request the scaleform movie
-    
+
     if not scaleform then -- If the scaleform is nil
         return error(('Failed to request scaleform movie - [%s]'):format(name)) -- Error the failed scaleform request
     end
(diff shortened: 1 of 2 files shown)


modules/scaleform.lua
```diff
@@ -0,0 +1,208 @@
...
+                    SetScriptGfxDrawBehindPausemenu(true) -- allow it to draw behind pause menu
+                    SetScaleformFitRendertarget(self.handle, true)
+                end
+
+                if self.fullScreen then
+                    DrawScaleformMovieFullscreen(self.handle, 255, 255, 255, 255, 0)
+                else
+                    if not self.x or not self.y or not self.width or not self.height then
+                        error('Properties not set for scaleform') -- Error properties not set for scaleform
+                        DrawScaleformMovieFullscreen(self.handle, 255, 255, 255, 255, 0)
```
D4isDAVID, line 173: I'm not sure what you want this to do. An error would stop the code from running any further. Either change it to `lib.print.error` or keep the error but remove the next line since it won't run.

modules/scaleform.lua
```diff
@@ -0,0 +1,208 @@
...
+
+---@return nil
+---@description Dispose of the scaleform
+function Scaleform:Dispose()
+    if self.handle then -- If the handle exists
+        SetScaleformMovieAsNoLongerNeeded(self.handle) -- Set the scaleform movie as no longer needed
+    end
+
+    if self.renderTarget then -- If the render target exists
+        ReleaseNamedRendertarget(self.targetName) -- Release the named render target
```
D4isDAVID, line 198: What if someone runs `RenderTarget` multiple times with different names? Do those render target names need releasing too?

modules/scaleform.lua
```diff
@@ -0,0 +1,208 @@
...
+---@field draw boolean
+---@field renderTarget number
+---@field targetName string
+---@field handle number
+local Scaleform = lib.class('Scaleform')
+
+---@param name string
+---@return nil
+---@description Create a new scaleform class
+function Scaleform:constructor(name)
```
D4isDAVID, line 13: We should let users provide everything with a table of arguments.
"Everything" being the name, fullscreen, and properties.

modules/scaleform.lua
```diff
@@ -0,0 +1,208 @@
...
+            error(('Unsupported Parameter type [%s]'):format(type(arg))) -- Error unsupported type
+        end
+    end
+    EndScaleformMovieMethod() -- End the scaleform movie method
+end
+
+---@param name string
+---@return nil
+---@description Request a scaleform method with no return value or parameters
+function Scaleform:Method(name)
```
D4isDAVID, line 62: Why is there 3 different methods for running a scaleform method? Can we not merge it into one function that accepts a table of arguments, and decide what to do based on the arguments?

modules/scaleform.lua
```diff
@@ -0,0 +1,208 @@
...
+    self.width = width
+    self.height = height
+end
+
+
+---@param name string
+---@param model string|number
+---@return nil
+---@description Create a render target for the scaleform - optional , only if you want to render the scaleform in 3D
+function Scaleform:RenderTarget(name, model)
```
D4isDAVID, line 130: Should this not be merged into `Draw`?

modules/scaleform.lua
```diff
@@ -0,0 +1,208 @@
...
+
+        self.renderTarget = GetNamedRendertargetRenderId(name) -- Get the named render target render id
+        self.targetName = name -- Set the target name
+    end
+end
+
+---@param shouldDraw boolean
+---@return nil
+---@description Draw the scaleform
+function Scaleform:Draw(shouldDraw)
```
D4isDAVID, line 150: I believe it would be more intuitive if there was one `startDrawing` method, and one `stopDrawing` method, or something similar to that nature. Moreover, is it a possibility that one would want to draw a scaleform multiple times? For example, multiple render targets? In such a case it would be nice to have a way to handle that, where you can decide where to start/stop drawing without affecting other draw threads.

modules/scaleform.lua
```diff
@@ -0,0 +1,208 @@
...
+                        error('Properties not set for scaleform') -- Error properties not set for scaleform
+                        DrawScaleformMovieFullscreen(self.handle, 255, 255, 255, 255, 0)
+                    else
+                        DrawScaleformMovie(self.handle, self.x, self.y, self.width, self.height, 255, 255, 255, 255, 0)
+                    end
+                end
+
+                if self.renderTarget then -- If the render target is true
+                    SetTextRenderId(1) -- Reset the text render id
+                end
```
D4isDAVID, line 181: Is this required for stuff rendered after this scaleform? It would be nice if the comments explained the purpose rather than what every single line is doing, when the code is fairly readable already.

modules/scaleform.lua
```diff
@@ -0,0 +1,208 @@
...
+    end
+
+    BeginScaleformMovieMethod(self.handle, name) -- Begin the scaleform movie method
+    local result = EndScaleformMovieMethodReturnValue() -- End the scaleform movie method with a return value
+
+    local timeout = 0
+    while not IsScaleformMovieMethodReturnValueReady(result) do  -- While the return value is not ready
+        Wait(0) -- Wait 0
+        timeout = timeout + 1 -- Increment the timeout
+        if timeout > 1000 then -- If the timeout is greater than 1000
```
D4isDAVID, line 87: We already require ox_lib for this, so use `lib.waitFor`.

D4isDAVID (changes requested):

modules/scaleform.lua
Mycroft-Studios: Mainly for easier reading. so you can easily tell the difference between methods. however, method and method args could be merged together, Return would have to be seperate though, since it works completely differently

modules/scaleform.lua
Mycroft-Studios: yes, otherwise things afterwards would be trying to render upon the current target, so, it sets the Id, draws what it needs to and then changes it back

modules/scaleform.lua
D4isDAVID: From what I understand, it is technically possible to provide parameters *and* receive a return value. That's mainly my concern and why I want to merge the three.

Mycroft-Studios: it would be quite messy to merge return into it and will 100% be better as a separate function, since the functions are different, adding args as a optional param to method could be good tho

modules/scaleform.lua
Mycroft-Studios: in which way?

modules/scaleform.lua
Mycroft-Studios: for the first point, as you can probably tell, my base handler, which i modified for this PR, is done that way, I Wasn't entirely sure on how it would look in a non-PoC project, hence it not being present, that is something i can add in tho.
for the second one, we would quite literally, be drawing it multiple times, i dont entirely see any scenario, where you would need to draw the exact same scaleform, on two different models, at the exact same time

D4isDAVID: Yes, for the former point it would be nice. As for the latter, if you don't see a use case for it, then no need to implement it.

modules/scaleform.lua
D4isDAVID: I'm not very knowledgeable on scaleforms, so I'm unaware if there's a use case where multiple render targets would be used, and in what way. Currently it's built where you setup a render target if you want to, and only that one is used, so I was thinking if it's a one-time thing where you only set up one render target, you can simply set it up in the beginning of the draw method and clean it up when done. If there's a use case for multiple render targets then it's possible to keep this method and also add it as an option to the constructor so you can create a render target immediately.

Manason (changes requested):
I'm confused by this API. I think it would help if all the methods of this class have a leading verb in the name. I also don't think this API is abstracting away enough of the lower level complexities of scaleform. Focus the API on how the user wants to interact with scaleform, and hide the way it actually works.

For an example API that does do some abstractions, see this C# scaleform class https://github.com/citizenfx/fivem/blob/master/code/client/clrcore/External/Scaleform.cs

Mycroft-Studios:
> I also don't think this API is abstracting away enough of the lower level complexities of scaleform

may i ask how see: 
https://cdn.discordapp.com/attachments/1012753554072674398/1288912766802989067/image.png?ex=66f6e90e&is=66f5978e&hm=16f102e273bf36f2c817350d7a7a11977016fed0686ecb10231e28f8328ba4dc&

for how this looks in code. it cannot get more simplier while still giving all scaleform functionality to the user

modules/scaleform.lua
```diff
@@ -0,0 +1,224 @@
+---@class Scaleform : OxClass
+---@field name string
+---@field scaleform number
+---@field draw boolean
+---@field target number
+---@field targetName string
+---@field handle number
+local Scaleform = lib.class('Scaleform')
+
+---@param details table | string
```
D4isDAVID, line 10: We should provide an actual type for the details table.

modules/scaleform.lua
```diff
@@ -0,0 +1,224 @@
...
+---@return nil
+---@description Create a new scaleform class
+function Scaleform:constructor(details)
+    local detailProvided = type(details) == "table"
+    self.name = detailProvided and details.name or details -- Set the name
```
D4isDAVID, line 15: `name` is only used to create the `handle`, is there any benefit to not keeping this as just a local variable?

modules/scaleform.lua
```diff
@@ -0,0 +1,224 @@
...
+        elseif type(arg) == 'boolean' then -- If the type of v is a boolean
+            ScaleformMovieMethodAddParamBool(arg) -- Add the boolean
+        else
+            error(('Unsupported Parameter type [%s]'):format(type(arg))) -- Error unsupported type
+        end
+    end
+end
+
+---@param type string
+---@return boolean | int | string
```
D4isDAVID, line 59:
```suggestion
---@param type 'bool' | 'int' | 'string'
---@return boolean | integer | string
```
Specific types are appreciated, makes it more readable.

modules/scaleform.lua
```diff
@@ -0,0 +1,224 @@
...
+local function retrieveReturnValue(type)
+    local result = EndScaleformMovieMethodReturnValue() -- End the scaleform movie method with a return value
+
+    lib.waitFor(function()
+        if IsScaleformMovieMethodReturnValueReady(result) then
+            return true
+        end
+    end, "Failed to retrieve return value", 1000)
+
+    if type == "int" then -- If the type is an integer
```
D4isDAVID, line 70: I think we should change it to accept Lua-like type strings, such as `'integer'` and `'boolean'`.

modules/scaleform.lua
```diff
@@ -0,0 +1,224 @@
...
+    if args then
+        convertArgs(args)
+    end
+
+    -- When wanting a return value, the scaleform has to be ended with a different native
+    if returnValue then
+        retrieveReturnValue(returnValue)
+    else
+        EndScaleformMovieMethod() -- End the scaleform movie method
+    end
```
D4isDAVID, line 105:
```suggestion
    if returnValue then
        return retrieveReturnValue(returnValue)
    end

    EndScaleformMovieMethod() -- End the scaleform movie method
```
Let's actually return the retrieved return value.

modules/scaleform.lua
```diff
@@ -0,0 +1,224 @@
...
+        return GetScaleformMovieMethodReturnValueString(result) -- Get the return value as a string
+    end
+end
+
+---@param name string
+---@param args? table
+---@param returnValue? string
+---@return any
+---@description Call a scaleform function, with optional args or return value.
+function Scaleform:method(name, args, returnValue)
```
D4isDAVID, line 84: Can be renamed to something like `callMethod` to have a verb in the beginning.

modules/scaleform.lua
```diff
@@ -0,0 +1,224 @@
...
+    self.y = y
+    self.width = width
+    self.height = height
+end
+
+---@param name string
+---@param model string|number
+---@return nil
+---@description Create a render target for the scaleform - optional , only if you want to render the scaleform in 3D
+function Scaleform:renderTarget(name, model)
```
D4isDAVID, line 135: Can be renamed to something like `setRenderTarget` to have a verb in the beginning.

D4isDAVID (changes requested):

ChatDisabled:
This got added to https://github.com/overextended/ox_lib/pull/698 instead. Unfortunately without proper attribution. If you however want to have proper attribution we can resolve this in DMs.

PR #588, opened 2024-09-29
fix(server/loops): correctly skip if society is too low

solareon:
## Description

Fixes erroneous issue where if society account has zero dollars it will still issue a paycheck

## Checklist

### DIFF
--- a/server/loops.lua
+++ b/server/loops.lua
@@ -32,7 +32,7 @@ local function pay(player)
         return
     end
     local account = config.getSocietyAccount(job.name)
-    if not account or account == 0 then -- Checks if player is employed by a society
+    if not account then -- Checks if player is employed by a society
         config.sendPaycheck(player, payment)
         return
     end


server/loops.lua
```diff
@@ -32,7 +32,7 @@ local function pay(player)
         return
     end
     local account = config.getSocietyAccount(job.name)
-    if not account or account == 0 then -- Checks if player is employed by a society
+    if not account then -- Checks if player is employed by a society
```
Manason, line 35: Would this be better as a config option? ie. pay even if account is empty?

Manason (approved):

server/loops.lua
mafewtm: I don't think we should be natively giving people the ability to use the society payment and still get paid even if the account doesn't have money. That completely negates the purpose of societal account payments.

mafewtm (approved):

PR #593, opened 2024-10-08
Use external state management

abdel1touimi:
## Checklist

### DIFF
--- a/bridge/qb/client/events.lua
+++ b/bridge/qb/client/events.lua
@@ -1,4 +1,5 @@
 local config = require 'config.client'
+local useExternalPlayerState = require 'config.shared'.useExternalPlayerState
 
 -- Trigger Command
 --- @deprecated
@@ -23,6 +24,8 @@ RegisterNetEvent('QBCore:Client:VehicleInfo', function(info)
 
     TriggerEvent('QBCore:Client:'..info.event..'Vehicle', data)
 end)
+
+if useExternalPlayerState then return end
 
 AddStateBagChangeHandler('hunger', ('player:%s'):format(cache.serverId), function(_, _, value)
     TriggerEvent('hud:client:UpdateNeeds', value, LocalPlayer.state.thirst)
--- a/client/loops.lua
+++ b/client/loops.lua
@@ -1,3 +1,6 @@
+local useExternalPlayerState = require 'config.shared'.useExternalPlayerState
+if useExternalPlayerState then return end
+
 local statusInterval = require 'config.client'.statusIntervalSeconds
 local playerState = LocalPlayer.state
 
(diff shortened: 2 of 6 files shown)


Manason (commented):
Would setting the hunger and thirst decay to 0 accomplish the same thing as this PR?

Manason:
> > Would setting the hunger and thirst decay to 0 accomplish the same thing as this PR?
> 
> No, because:
> 
>     1. **`bridge/qb/client/events.lua`**:
>        Statebag handlers for thirst, hunger, and stress will still be present. While these primarily update the HUD, it would still be beneficial to remove them to clean things up.
> 
>     2. **`client/loops.lua`**:
>        Player health will continue to decrease despite setting the decay to 0.
> 
>     3. **`server/events.lua`**:
>        Statebag handlers will still be active, syncing player metadata with the player state for hunger and thirst.
> 
>     4. **`server/loops.lua`**:
>        Even if decay is set to 0, metadata updates will continue.
> 
>     5. **`server/player.lua`**:
>        It’s redundant at this point. I see no need to keep it, but that’s just my opinion.

One of my concerns is breaking our existing API behavior. Other resources set metadata or a statebag and expect it to be synced and updated. Maybe dropping thirst/hunger should be moved to config rather than being hard coded. I'd like to find a way to add compatibility for your resource without changing the behavior for existing Qbox integrations.

abdel1touimi:
> > > Would setting the hunger and thirst decay to 0 accomplish the same thing as this PR?
> > 
> > 
> > No, because:
> > ```
> > 1. **`bridge/qb/client/events.lua`**:
> >    Statebag handlers for thirst, hunger, and stress will still be present. While these primarily update the HUD, it would still be beneficial to remove them to clean things up.
> > 
> > 2. **`client/loops.lua`**:
> >    Player health will continue to decrease despite setting the decay to 0.
> > 
> > 3. **`server/events.lua`**:
> >    Statebag handlers will still be active, syncing player metadata with the player state for hunger and thirst.
> > 
> > 4. **`server/loops.lua`**:
> >    Even if decay is set to 0, metadata updates will continue.
> > 
> > 5. **`server/player.lua`**:
> >    It’s redundant at this point. I see no need to keep it, but that’s just my opinion.
> > ```
> 
> One of my concerns is breaking our existing API behavior. Other resources set metadata or a statebag and expect it to be synced and updated. Maybe dropping thirst/hunger should be moved to config rather than being hard coded. I'd like to find a way to add compatibility for your resource without changing the behavior for existing Qbox integrations.

I understand the issue.
I will implement some change in the player states and be back to you

abdel1touimi:
change done in playerstates.

Now it accept changes done directly in metadata or statebag

PR #597, opened 2024-10-19
Pause Menu Event

AfterLifeStudio:
i was working on the hud and thought this event can be good in the core cause the dev will not have to create the pause menu logic in every script he want to use and also because i have seen it in other frameworks so i thought this could be a nice little pull request in the core, let me know if we can have something similar or anything like this in the core


client/loops.lua
```diff
@@ -1,5 +1,6 @@
 local statusInterval = require 'config.client'.statusIntervalSeconds
 local playerState = LocalPlayer.state
+local Active = false
```
mafewtm, line 3: The `Active` variable isn't really necessary as `IsPauseMenuActive()` already returns a boolean value. Additionally, local variables should be camelCase.

Manason (commented):
Will multiple resources be consuming this event? What are some of the use cases?

client/loops.lua
Manason: The variable is needed to capture the old state to understand if there is a change.

JGUsman007:
yes multiple resources will be using this event to check if the pausemenu is active so they can hide the ui on the screen or to cancel some action the player is performing

JGUsman007:
> Will multiple resources be consuming this event? What are some of the use cases?

yes multiple resources will be using this event to check if the pausemenu is active so they can hide the ui on the screen or to cancel some action the player is performing

Manason:
> > Will multiple resources be consuming this event? What are some of the use cases?
> 
> yes multiple resources will be using this event to check if the pausemenu is active so they can hide the ui on the screen or to cancel some action the player is performing

Why should core own the pause menu event rather than hud?

JGUsman007:
> > > Will multiple resources be consuming this event? What are some of the use cases?
> > 
> > 
> > yes multiple resources will be using this event to check if the pausemenu is active so they can hide the ui on the screen or to cancel some action the player is performing
> 
> Why should core own the pause menu event rather than hud?

because its not just for hiding ui its for other stuff like to cancel the player action it opens pausemenu or just not hud ui for other ui elements in server too to check if the player opens the pausemenu so it just hide the ui,and if the hud handles this event all the other scripts that are using this event not just to hide ui but to limit actions or cancel the action if it gets activated will be dependent on the hud resource

client/loops.lua
```diff
@@ -15,3 +16,18 @@ CreateThread(function()
...
+        if PauseMenuActive and not Active then
+            Active = true
+            TriggerEvent('qbx_core:client:onPauseMenuActive', Active)
+        elseif Active and PauseMenuActive then
+            Active = false
+            TriggerEvent('qbx_core:client:onPauseMenuActive', Active)
+        end
+        Wait(1000)
+    end
+end)
```
Manason, line 33: Use https://overextended.dev/ox_lib/Modules/Cache/Shared#adding-new-cached-values and lib.onCache to do the event handling.

Manason (changes requested):

client/loops.lua
mafewtm: I don't think they'll add a new cache option for something that's just true or false. I don't see a real use for us making a new cached value for it either

Issue #577, opened 2024-09-20
Create User Object


Manason:
### The problem

- Different FiveM accounts can't play on the same copy of GTA because user accounts are tied to license instead of FiveM Id
- User accounts are stamped down onto each character, which duplicates the data

### Ideal solution

- Introduce a new Users table and extract the fields common between all characters such as license2
- Introduce an auto incrementing UserId
- FK the UserId on each character to associate to a user
- Associate each user to a FiveM Id

PR #598, opened 2024-10-19
feat(server): user object

mafewtm:
## Description

Creates an entry into the `users` table upon connection. It autoincrements and logs the following identifiers:

- username
- license
- license2
- fivem
- discord

Doesn't completely implement #577 yet but this seems like a good starting point that doesn't break anything.

## Checklist


qbx_core.sql
```diff
@@ -1,3 +1,13 @@
+CREATE TABLE IF NOT EXISTS `users` (
+  `userId` int UNSIGNED NOT NULL AUTO_INCREMENT,
+  `username` varchar(50) DEFAULT NULL,
```
Manason, line 3: Should we store this info? This doesn't seem like a useful field

Manason (approved):
Probably also need some conversion SQL code to initialize the users table. Consider baking that into server startup to do if the users table is empty.

qbx_core.sql
mafewtm: It's there so you can see whos who in the db. Otherwise its just a bunch of numbers and letters. It's also easier to filter the DB by the name if you don't know the license. We have it in the players table with all their characters as well.

mafewtm:
> Probably also need some conversion SQL code to initialize the users table. Consider baking that into server startup to do if the users table is empty.

I'm assuming you mean having it autorun the users table instead of how we have all the resources currently where they run their own sql?

Manason:
> > Probably also need some conversion SQL code to initialize the users table. Consider baking that into server startup to do if the users table is empty.
> 
> I'm assuming you mean having it autorun the users table instead of how we have all the resources currently where they run their own sql?

I think so? I just mean running the SQL via lua rather than requiring users to manually run a SQL file to convert things.

mafewtm:
> > > Probably also need some conversion SQL code to initialize the users table. Consider baking that into server startup to do if the users table is empty.
> > 
> > 
> > I'm assuming you mean having it autorun the users table instead of how we have all the resources currently where they run their own sql?
> 
> I think so? I just mean running the SQL via lua rather than requiring users to manually run a SQL file to convert things.

Well nothing needs to be converted. This is brand new and doesn't change any preexisting code. Nevertheless, I can do it if that's what we wanna do? Should probably do it for all the SQL then if we're gonna do it for this

Manason:
> > > > Probably also need some conversion SQL code to initialize the users table. Consider baking that into server startup to do if the users table is empty.
> > > 
> > > 
> > > I'm assuming you mean having it autorun the users table instead of how we have all the resources currently where they run their own sql?
> > 
> > 
> > I think so? I just mean running the SQL via lua rather than requiring users to manually run a SQL file to convert things.
> 
> Well nothing needs to be converted. This is brand new and doesn't change any preexisting code. Nevertheless, I can do it if that's what we wanna do? Should probably do it for all the SQL then if we're gonna do it for this

Well, it only gets inserted on connection. So as the code is currently written, this table can't be relied upon to do operations across all users, only ones that have logged in since upgrading.

mafewtm:
> > > > > Probably also need some conversion SQL code to initialize the users table. Consider baking that into server startup to do if the users table is empty.
> > > > 
> > > > 
> > > > I'm assuming you mean having it autorun the users table instead of how we have all the resources currently where they run their own sql?
> > > 
> > > 
> > > I think so? I just mean running the SQL via lua rather than requiring users to manually run a SQL file to convert things.
> > 
> > 
> > Well nothing needs to be converted. This is brand new and doesn't change any preexisting code. Nevertheless, I can do it if that's what we wanna do? Should probably do it for all the SQL then if we're gonna do it for this
> 
> Well, it only gets inserted on connection. So as the code is currently written, this table can't be relied upon to do operations across all users, only ones that have logged in since upgrading.

Correct. That's how it gets the information. We don't have all the information available to fill out so as they login, they get a database entry.

D4isDAVID (approved):

PR #603, opened 2024-10-21
ADD IF NOT EXISTS syntax error during installation from TXAdmin Recipe

daixso:
## Description

Fixes a syntax error preventing proper database setup during installation from TXAdmin recipe. IF NOT EXISTS is improper syntax after ADD you only need the ADD and then the row you wish to add.

## Checklist


D4isDAVID:
Please install a MariaDB server instead of a regular MySQL server. This syntax does exist there, and the purpose is for people migrating from older versions where this column wouldn't exist.

daixso:
Ah interesting I will try that and close if that fixes thank you!

PR #607, opened 2024-10-24
fix(client/config): adjust to match ox_lib

mafewtm:
## Description

With the new default timeout being increased to 30 seconds in ox_lib, our core config option should reflect that.

I'm also of the opinion that it can now be removed but that's up to yall.

## Checklist


Manason (approved):
I also agree that it can (and should) be removed.

D4isDAVID (approved):

Issue #608, opened 2024-10-24
qbx_core:GetOfflinePlayer Runtime Error - not handling nil gracefully


zozomanx:
### Summary

When passing a Citizen ID that doesn't exist to GetOfflinePlayer a runtime error occurs

### Reproduction

1. Enter in an non-existant Citizen ID into exports.qbx_core:GetOfflinePlayer() - Use my provided commands to test
2. Attempt to run that export via the command
3. Get error

### Expected behavior

Be returned a nil value if that citizen ID doesn't exist or have it be handled the same way exports.qbx_core:GetPlayerByCitizenId() does.

exports.qbx_core:GetPlayer() and exports.qbx_core:GetPlayerByCitizenId() handle an incorrect CID or Citizen ID gracefully and do not generate runtime errors.

### Actual behavior

A runtime error occurs and the program exits.

### Additional context

**This is the runtime error that gets generated**:

[     script:rfl_core] command 'gunband' failed to execute!
[     script:rfl_core] @rfl_core/server/server.lua:95:
[     script:rfl_core]  An error occurred while calling export `GetOfflinePlayer` in resource `qbx_core`:
[     script:rfl_core]   citizen:/scripting/lua/scheduler.lua:739: SCRIPT ERROR: @qbx_core/server/storage/players.lua:115: attempt to index a nil value (local 'player')
[     script:rfl_core]   > fetchPlayerEntity (@qbx_core/server/storage/players.lua:115)
[     script:rfl_core]   > ref (@qbx_core/server/player.lua:57)
[     script:rfl_core]   > rawQuery (@oxmysql/dist/build.js:27077)
[     script:rfl_core]   > processTicksAndRejections (node:internal/process/task_queues:96)

**Commands I used to debug with and you can easily copy and paste**:

```
lib.addCommand('niltestid', {
    help = 'Test command',
    params = {
        { name = 'id', type = 'string' },
    }
}, function(source, args)
    local player = exports.qbx_core:GetPlayer(args.id)
    print(player)
    if not player then
        TriggerClientEvent('ox_lib:notify', source, { title = 'Error', description = 'Invalid argument', position = 'top', type = 'error' })
        return
    end
end)

lib.addCommand('niltestcid', {
    help = 'Test command',
    params = {
        { name = 'id', type = 'string'},
    }
}, function(source, args)
    local player = exports.qbx_core:GetPlayerByCitizenId(args.id)
    print(player)
    if not player then
[...]

PR #610, opened 2024-10-24
fix: Add guard clause to fetchPlayerEntity

zozomanx:
## Description

Added guard clause to fetchPlayerEntity
This should be merged because when given an incorrect citizenid the function wouldn't fail gracefully and would cause a runtime error. Now nil will be returned which can be handled.

Yes, it fixes an issue.

## Checklist


D4isDAVID:
Is this about your issue #608? Please read my response; that issue was already fixed in the latest release. Please update your qbx_core.

zozomanx:
> Is this about your issue #608? Please read my response; that issue was already fixed in the latest release. Please update your qbx_core.

I somehow missed your reply in there, sorry for this request. I just tested the latest release and it seems to be good. Thank you!

PR #611, opened 2024-10-26
feat(server): deprecate player functions

mafewtm:
## Description

This PR deprecated all `player.Functions...` in favor of having some of them be exports. A list of their replacements are below:

_Identifier is either source or citizenId as to handle both online and offline players._

**Replacements:**
```lua
player.Functions.SetJob(jobName, grade) → exports.qbx_core:SetJob(identifier, jobName, grade)
player.Functions.SetGang(gangName, grade) → exports.qbx_core:SetGang(identifier, gangName, grade)
player.Functions.SetJobDuty(onDuty) → exports.qbx_core:SetJobDuty(identifier, onDuty)
player.Functions.SetMetaData(meta, val) → exports.qbx_core:SetMetadata(identifier, metadata, value)
player.Functions.GetMetaData(meta) → exports.qbx_core:GetMetadata(identifier, metadata)
player.Functions.AddMoney(moneytype, amount, reason) → exports.qbx_core:AddMoney(identifier, moneyType, amount, reason)
player.Functions.RemoveMoney(moneytype, amount, reason) → exports.qbx_core:RemoveMoney(identifier, moneyType, amount, reason)
player.Functions.SetMoney(moneytype, amount, reason) → exports.qbx_core:SetMoney(identifier, moneyType, amount, reason)
player.Functions.GetMoney(moneytype) → exports.qbx_core:GetMoney(identifier, moneyType)
```

**Removals:**
```lua
player.Functions.SetCreditCard(cardNumber) -- use SetCharInfo instead
player.Functions.AddJobReputation(amount) -- use SetMetadata instead
player.Functions.Functions.Save() -- use Save or SaveOffline instead
```

**Additions:**
```lua
exports.qbx_core:SetCharInfo(identifier, charInfo, value)
```

I've tested most of these and haven't encountered any issues. Idk if any of this should be done different or if we want to add/remove anything else.

## Checklist


config/server.lua
```diff
@@ -127,7 +127,7 @@ return {
     ---@param player Player Player object
     ---@param payment number Payment amount
     sendPaycheck = function (player, payment)
-        player.Functions.AddMoney('bank', payment)
+        AddMoney(player.PlayerData.source, 'bank', payment)
```
Manason, line 130: This needs to be an export call, no? Since it will be invoked from within another script, the AddMoney function won't exist in the other script.

Manason (approved):

server/player.lua
```diff
@@ -239,12 +314,64 @@ end
...
+        end
+    end
+
+    local success, errorResult = SetPlayerPrimaryGang(player.PlayerData.citizenid, gangName)
+
+    if not success then
+        return false, errorResult
+    end
+
+    return true
```
D4isDAVID, line 364:
```suggestion
    return SetPlayerPrimaryGang(player.PlayerData.citizenid, gangName)
```
I think we can simplify this section.

server/player.lua
```diff
@@ -61,6 +61,78 @@ end
...
+        end
+    end
+
+    local success, errorResult = SetPlayerPrimaryJob(player.PlayerData.citizenid, jobName)
+
+    if not success then
+        return false, errorResult
+    end
+
+    return true
```
D4isDAVID, line 112:
```suggestion
    return SetPlayerPrimaryJob(player.PlayerData.citizenid, jobName)
```
Ditto

server/player.lua
```diff
@@ -239,12 +314,64 @@ end
 
 exports('RemovePlayerFromJob', RemovePlayerFromJob)
 
+---Removes the player from their current primary gang and adds the player to the new gang
+---@param identifier Source | string
+---@param grade? integer defaults to 0
+---@return boolean success if gang was set
+---@return ErrorResult? errorResult
```
D4isDAVID, line 321:
```suggestion
---@param identifier Source | string
---@param gangName string
---@param grade? integer defaults to 0
---@return boolean success if gang was set
---@return ErrorResult? errorResult
```
Missing `gangName`

server/player.lua
```diff
@@ -580,259 +710,101 @@ function CreatePlayer(playerData, Offline)
     self.PlayerData = playerData
     self.Offline = Offline
 
+    ---@deprecated use UpdatePlayerData instead
     function self.Functions.UpdatePlayerData()
-        if self.Offline then return end -- Unsupported for Offline Players
-        TriggerEvent('QBCore:Player:SetPlayerData', self.PlayerData)
-        TriggerClientEvent('QBCore:Player:SetPlayerData', self.PlayerData.source, self.PlayerData)
+        assert(not self.Offline, 'unsupported for offline players')
```
D4isDAVID, line 715: This is pretty much a breaking change as it will stop the caller's code from being executed, when previously it would continue. Consider switching this with a `lib.print.warn`.

server/player.lua
```diff
@@ -1060,6 +1036,281 @@ end
...
+
+    TriggerEvent('QBCore:Player:SetPlayerData', player.PlayerData)
+    TriggerClientEvent('QBCore:Player:SetPlayerData', player.PlayerData.source, player.PlayerData)
+end
+
+---@param identifier Source | string
+---@param metadata string
+---@param value any
+function SetMetadata(identifier, metadata, value)
+    if not metadata or type(metadata) ~= 'string' then return end
```
D4isDAVID, line 1068:
```suggestion
    if type(metadata) ~= 'string' then return end
```
Why not simplify to this? Same with some other functions.

server/player.lua
```diff
@@ -1060,6 +1036,281 @@ end
...
+---@param charInfo string
+---@param value any
+function SetCharInfo(identifier, charInfo, value)
+    if not charInfo or type(charInfo) ~= 'string' then return end
+
+    local player = type(identifier) == 'string' and (GetPlayerByCitizenId(identifier) or GetOfflinePlayer(identifier)) or GetPlayer(identifier)
+
+    if not player then return end
+
+    --local oldCharInfo = player.PlayerData.charinfo[charInfo]
```
D4isDAVID, line 1135: Will this be used for something?

D4isDAVID (changes requested):

server/player.lua
mafewtm: This was kept this way so that the SetGang player function would still operate the same way.

server/player.lua
mafewtm: Ditto

server/player.lua
mafewtm: depends if we want to add something like `qbx_core:server:onSetCharInfo`

server/player.lua
D4isDAVID: It would still operate the same way, but ok.

mafewtm: I just did what I was told. Maybe @Manason can explain more.

Manason: I agree with d4's suggestion

server/player.lua
```diff
@@ -1060,6 +1028,281 @@ end
 
 exports('SaveOffline', SaveOffline)
 
+---@param identifier Source | string
+---@param key string
+---@param value any
+function SetPlayerData(identifier, key, value)
+    if not key or type(key) ~= 'string' then return end
```
D4isDAVID, line 1035:
```suggestion
    if type(key) ~= 'string' then return end
```
Can be simplified

server/player.lua
```diff
@@ -1060,6 +1028,281 @@ end
...
+    end
+end
+
+exports('SetMetadata', SetMetadata)
+
+---@param identifier Source | string
+---@param metadata string
+---@return any
+function GetMetadata(identifier, metadata)
+    if not metadata or type(metadata) ~= 'string' then return end
```
D4isDAVID, line 1106:
```suggestion
    if type(metadata) ~= 'string' then return end
```

server/player.lua
```diff
@@ -1060,6 +1028,281 @@ end
...
+    return player.PlayerData.metadata[metadata]
+end
+
+exports('GetMetadata', GetMetadata)
+
+---@param identifier Source | string
+---@param charInfo string
+---@param value any
+function SetCharInfo(identifier, charInfo, value)
+    if not charInfo or type(charInfo) ~= 'string' then return end
```
D4isDAVID, line 1121:
```suggestion
    if type(charInfo) ~= 'string' then return end
```

D4isDAVID (changes requested):

D4isDAVID (approved):

PR #612, opened 2024-10-27
fix(server/player) sanitize player name & limit to 50 chars for db constraints

blaaaak:
fix(server/player) sanitize player name, limit to 50 chars for db constraints

## Description

- Fixes when players with super long names join and try to create a character, the name would be too long for the name column in the players table and cause errors. 
- Added basic string sanitization to remove any non A-Z 0-9 characters. 

Added a

## Checklist

### DIFF
--- a/server/player.lua
+++ b/server/player.lua
@@ -425,7 +425,7 @@ function CheckPlayerData(source, playerData)
     if source then
         playerData.source = source
         playerData.license = playerData.license or GetPlayerIdentifierByType(source --[[@as string]], 'license2') or GetPlayerIdentifierByType(source --[[@as string]], 'license')
-        playerData.name = GetPlayerName(source)
+        playerData.name = string.sub(GetPlayerName(source):gsub('[^%w%s]', ''), 1, 50)
         Offline = false
     end
 


server/player.lua
```diff
@@ -425,7 +425,7 @@ function CheckPlayerData(source, playerData)
     if source then
         playerData.source = source
         playerData.license = playerData.license or GetPlayerIdentifierByType(source --[[@as string]], 'license2') or GetPlayerIdentifierByType(source --[[@as string]], 'license')
-        playerData.name = GetPlayerName(source)
+        playerData.name = string.sub(GetPlayerName(source):gsub('[^%w%s]', ''), 1, 50)
```
mafewtm, line 428: The DB allows names to be 255 characters as its set to `varchar(255)`. I don't think the player name should be touched what so ever in my opinion.

blaaaak: image
Do I have an old version of the SQL, this was ran from the qbx_core resource, sets the "name" to 50.

blaaaak: I do see this now, my apologies not entirely sure why the database set it to 50 for me.

blaaaak: nvm I missed this, the player table is altered to set 'name' to `varchar(50)`

https://github.com/Qbox-project/qbx_core/blob/main/qbx_core.sql#L24-L25

PR #614, opened 2024-10-31
fix(shared): clear deprecated tables

mafewtm:
## Description

Clears the deprecated shared tables in order to prevent confusion. Currently a common question is `"Why doesn't my armor work?"`. It's asked not knowing that the two old QB armors are no longer supported but are still in the items table.

## Checklist


mafewtm:
If the items table is a worry, we could add something like:
```lua
---@deprecated This file is deprecated and will be removed in the future. Please add your items directly in ox_inventory/data/items.lua file. Currently items placed in here will be converted at next server restart.
---@type table<string, Item>
return exports.ox_inventory:Items()
```

Manason (commented):
If we're going to delete these, I think we need to also replace the default ox_inventory data files in the recipe to contain the same items

mafewtm:
> If we're going to delete these, I think we need to also replace the default ox_inventory data files in the recipe to contain the same items

The reason to delete these is because the vast majority of these items are unused. I mean if its absolutely necessary we could create a default list of whatever.

Manason (approved):

Manason:
> > If we're going to delete these, I think we need to also replace the default ox_inventory data files in the recipe to contain the same items
> 
> The reason to delete these is because the vast majority of these items are unused. I mean if its absolutely necessary we could create a default list of whatever.

Ok sounds good. Just need to be mindful as we rely on items in other scripts to add them to the recipe

D4isDAVID (approved):

PR #617, opened 2024-11-02
refactor: enhance draw text functiones for better flexibility

GoldO-dev:
## Description

- Changed `scale` parameter in `qbx.drawText3d` to be a `vec2` with a default value of `vec2(0.35, 0.35)`, while maintaining backwards compatibility with the previous version
- Introduced `disableDropShadow` and `disableOutline` parameters in `qbx.drawText2d`
- Introduced `enableDropShadow` and `enableOutline` parameters in `qbx.drawText3d`
- Added a warning for users using a number value as the scale parameter in `qbx.drawText3d`, encouraging the use of a `vec2` instead.

## Checklist

### DIFF
--- a/modules/lib.lua
+++ b/modules/lib.lua
@@ -1,4 +1,5 @@
 local isServer = IsDuplicityVersion()
+local hasText3dScaleAsNumberDepricatedMessageShown = false
 
 local qbx = {}
 qbx.string = {}
@@ -330,6 +331,8 @@ else
     ---@field coords vector2
     ---@field width? number default: `1.0`
     ---@field height? number default: `1.0`
+    ---@field disableDropShadow? boolean
+    ---@field disableOutline? boolean
 
     ---Draws text onto the screen in 2D space for a single frame.
     ---@param params LibDrawText2DParams
@@ -341,12 +344,19 @@ else
         local color = params.color or vec4(255, 255, 255, 255)
         local width = params.width or 1.0
         local height = params.height or 1.0
+        local disableDropShadow = params.disableDropShadow or false
+        local disableOutline = params.disableOutline or false
 
         SetTextScale(scale, scale)
         SetTextFont(font)
         SetTextColour(math.floor(color.r), math.floor(color.g), math.floor(color.b), math.floor(color.a))
-        SetTextDropShadow()
-        SetTextOutline()
+        if not disableDropShadow then
+            SetTextDropShadow()
+        end
+        if not disableOutline then
+            SetTextOutline()
+        end
+
         SetTextCentre(true)
         BeginTextCommandDisplayText('STRING')
         AddTextComponentSubstringPlayerName(text)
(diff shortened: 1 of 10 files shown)


modules/lib.lua
```diff
@@ -356,19 +366,38 @@ else
...
+        local scale = (isScaleparamANumber and vec2(params.scale, params.scale))
+                  or params.scale
+                  or vec2(0.35, 0.35)
         local font = params.font or 4
         local color = params.color or vec4(255, 255, 255, 255)
+        local enableDropShadow = params.enableDropShadow or false
+        local enableOutline = params.enableOutline or false
+        if isScaleparamANumber and not hasText3dScaleAsNumberDepricatedMessageShown then
+            lib.print.warn('This resource is still using qbx.drawText3d with a number as the scale parameter!')
+            lib.print.warn('If you are the author, please update to use a vec2 for the scale parameter. If you are not, please tell them to update!')
```
Manason, line 388: Let's remove this. I don't think we need to force people to update their scripts over this.

modules/lib.lua
```diff
@@ -341,12 +344,19 @@ else
         local color = params.color or vec4(255, 255, 255, 255)
         local width = params.width or 1.0
         local height = params.height or 1.0
+        local disableDropShadow = params.disableDropShadow or false
+        local disableOutline = params.disableOutline or false
```
Manason, line 348: Why does 2D use disable while 3D uses enable? 2D defaults these features to on if unset whereas 3D does not.

Manason (changes requested):

modules/lib.lua
GoldO-dev: > Why does 2D use disable while 3D uses enable? 2D defaults these features to on if unset whereas 3D does not.

The reason I did it like that is because before I changed stuff there was a outline and drop shadow on the 2d text by default and none on the 3d text by default. So I wanted to keep it like that.  

This is the original code for drawText2d as you can see by default there is a outline and drop shadow
```lua
    function qbx.drawText2d(params)
        local text = params.text
        local coords = params.coords
        local scale = params.scale or 0.35
        local font = params.font or 4
        local color = params.color or vec4(255, 255, 255, 255)
        local width = params.width or 1.0
        local height = params.height or 1.0
        SetTextScale(scale, scale)
        SetTextFont(font)
        SetTextColour(math.floor(color.r), math.floor(color.g), math.floor(color.b), math.floor(color.a))
        SetTextDropShadow()
        SetTextOutline()
        SetTextCentre(true)
        BeginTextCommandDisplayText('STRING')
        AddTextComponentSubstringPlayerName(text)
        EndTextCommandDisplayText(coords.x - width / 2, coords.y - height / 2 + 0.005)
    end
```

modules/lib.lua
GoldO-dev: Done

modules/lib.lua
GoldO-dev: Do you want me to change it? Or can I keep it the way I've done?

GoldO-dev: @Manason am I allowed to mark this conversation as resolved?

Manason: Yeah let's default it either way for both consistently.

GoldO-dev: Not sure I totally understand, what you mena by that. Can you elaborate?

Manason: Let's either default shadow to on and have the disable param, or default it to off and have the enable param. But I don't want them to have different params.

GoldO-dev: Okay, what would you prefer?

Manason: Probably default off with an optional enable param

GoldO-dev: Okay, will change it quick. I also have a updated public documentation fork, where I will change the stuff on too. And make a pull request on the docs repository.

GoldO-dev:
@Manason all requested changes has been made.

I've also made a pull request to the docs repository that updates the docs to match the changes made in this pull request. You can take a look at it [here](https://github.com/Qbox-project/qbox-project.github.io/pull/43)

Manason (approved):

modules/lib.lua
```diff
@@ -356,19 +365,31 @@ else
     ---@class LibDrawText3DParams : LibDrawTextParams
     ---@field coords vector3
     ---@field disableDrawRect? boolean
+    ---@field scale? vector2 default: `vec2(0.35,0.35)`
```
Manason, line 368: Can you add back integer as well? I think we'll support both going forward rather than deprecate the original integer param.

Manason (approved):

modules/lib.lua
GoldO-dev: The way it's done at the moment it supports a integer as well. It just turns it into a vector2 under the hood since it make the code a bit easier to read. 

Would you like me to do it like this?
```lua
---@field scale? integer | vector2 default: `0.35`
```

Since the code already handles integers, I don't see a need to change any of the code in the actual function. Correct me if I'm wrong.

Manason: You are correct.

GoldO-dev: @Manason the latest commit should fix it. Can you check to confirm it's made how you wanted?

Manason (approved):

GoldO-dev:
@Manason just a quick question, do I need a second reviewer to review before it can be merged? Or can it be merged by just you giving it a review?

Manason:
> @Manason just a quick question, do I need a second reviewer to review before it can be merged? Or can it be merged by just you giving it a review?

A second reviewer is required.

mafewtm (approved):

PR #620, opened 2024-11-11
fix(server/player): typo

mafewtm:
## Description

Fixes a typo which has been present for 9 months I guess

## Checklist

### DIFF
--- a/server/player.lua
+++ b/server/player.lua
@@ -903,6 +903,7 @@ function CreatePlayer(playerData, Offline)
 
     AddEventHandler('qbx_core:server:onJobUpdate', function(jobName, job)
         if self.PlayerData.job.name ~= jobName then return end
+
         if not job then
             self.PlayerData.job = {
                 name = 'unemployed',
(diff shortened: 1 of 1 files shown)


server/player.lua
```diff
@@ -904,7 +904,7 @@ function CreatePlayer(playerData, Offline)
     AddEventHandler('qbx_core:server:onJobUpdate', function(jobName, job)
         if self.PlayerData.job.name ~= jobName then return end
         if not job then
-            self.Functions.setJob('unemployed', 0)
+            SetJob(self.PlayerData.source, 'unemployed', 0)
```
Manason, line 907: I wonder if this line should be deleted altogether?

mafewtm: Do you remember why you added it initially?

solareon: The gang version below doesn't appear to do the same set. I think these were from before qbx_management used exports for the multijob?

Manason: I'm not sure why this line is here. If the event is triggered, I assume it is because the player's job is already unemployed, so no need to set it again.

mafewtm: I've changed it to now mirror onGangUpdate.

solareon: To answer the purpose we listen to this event for the RemoveJob/AddJob exports so that other folks could integrate this into their scripts via the netevent.

solareon (approved):

Manason (approved):
Using SetJob() and SetGang() for the unemployed case worked fine I think. This way works too.

PR #621, opened 2024-11-13
feat(server): full vehicle persistence

CodexisPhantom:
## Description

This pull request will add the possibility to qbox users to enable a full vehicle persistence so that cars never despawn even after server restart. I tested this code with 10 people on a clean qbox server.

The convar `qbx:enableVehiclePersistence` has been modified :
- 0 : no persistence
- 1 : old persistence
- 2 : new persistence

## Checklist


server/vehicle-persistence.lua
```diff
@@ -128,4 +139,32 @@ AddEventHandler('entityRemoved', function(entity)
...
+        SetVehicleDoorsLocked(veh, 2)
+        SetEntityHeading(veh, vehicle.coords.w)
+    end
+end
+
+RegisterNetEvent('QBCore:Server:OnPlayerLoaded', function()
+    local players = exports.qbx_core:GetQBPlayers()
+    if spawned and #players ~= 1 then return end
+    spawnVehicles()
+    spawned = true
```
Manason, line 169: Won't this spawn the vehicles duplicate times every time a player joins?

Manason (approved):

Manason (changes requested):
Accidentally approved. Ignore this comment.

server/vehicle-persistence.lua
CodexisPhantom: Yep my bad I did some changes before push and I was sure it was right. I'll make a fix.

CodexisPhantom: I pushed the fix

server/vehicle-persistence.lua
```diff
@@ -128,4 +139,31 @@ AddEventHandler('entityRemoved', function(entity)
...
+        Entity(veh).state:set('vehicleid', vehicle.id, false)
+        SetVehicleDoorsLocked(veh, 2)
+        SetEntityHeading(veh, vehicle.coords.w)
+    end
+end
+
+RegisterNetEvent('QBCore:Server:OnPlayerLoaded', function()
+    local players = exports.qbx_core:GetQBPlayers()
+    if not spawned and #players == 1 then spawnVehicles() end
+end)
```
Manason, line 169: Would this work as an onResourceStart event? Why does there need to be at least 1 client in the server for this to work?

Manason: My understanding is that CreateVehicleServerSetter blocks waiting for a client to be in one sync range, so I suspect it should work fine if there are no clients on the server; it should simply wait for a client to come in range.

CodexisPhantom: Interesting, I'll test it now and push a modification if needed.

Manason: If we're going to precompute the vehicle type we should do so in shared/vehicles.lua

We'd just need to ensure that every car in the game is represented in that file, which would include any add-on vehicles that servers download.

What happens if you run the function with no players connected and then a player connects and enters one sync range of the vehicle spawn?

Manason: @solareon @mafewtm @D4isDAVID Any thoughts on this?

mafewtm: You're gonna have to explain what you want thoughts on

Manason: If we should add types to shared/vehicles

mafewtm: Why do we need types in the vehicles.lua? What purpose does it server that category, hash, or model does not or could not?

Manason: CreateVehicleServerSetter needs the type, so it requires us to have a client and spawn a vehicle to find the type. Category wouldn't work because servers use that field to categorize vehicles into sports cars etc

mafewtm: I mean if it *has* to be added then I guess. If that's what we wanna do then I'll do a pr. Just lmk.

CodexisPhantom: I can do it in this pr if you want

mafewtm: No I got it. I'm doing a couple more things to it anyway

Manason: Any chance you could write a script for it so that it's easy for me to update my config? lol

Manason: There's actually a pretty good argument for creating a separate table to hold this data so that people can easily upgrade. Otherwise merging a new field in with existing data is a big pain

CodexisPhantom: I was about to create the script and I thought the same but it's strange to have all data in a table and one data in another.

Manason: Our current state is we have config data mixed with non config data. So it makes sense to split out fields of shared/vehicles into data/vehicles.lua for fields that aren't meant to be opinionated

mafewtm: I'm not a particular fan of having two mile long data tables for the same thing either

CodexisPhantom: Thanks I'll modify my PR when I go back home from my vacation 😉

CodexisPhantom:
Hello guys,
Sorry for the time that I took to do this update but I finally did it. Can you check if it's good now @Manason @mafewtm.

server/vehicle-persistence.lua
```diff
@@ -121,4 +132,50 @@ AddEventHandler('entityRemoved', function(entity)
...
+    })
+    exports.qbx_core:EnablePersistence(veh)
+    Entity(veh).state:set('vehicleid', id, false)
+    SetVehicleDoorsLocked(veh, 2)
+    TriggerClientEvent('qbx_core:client:removeVehZone', -1, id)
+end
+
+lib.callback.register('qbx_core:server:getVehiclesToSpawn', function()
+    local vehicles = {}
+    local query = 'SELECT id, plate, coords FROM player_vehicles WHERE state = 0'
```
Manason, line 163: Rather than being queried by every player. This data should be queried on server startup and cached. Then, as vehicles get spawned, the cache gets smaller and smaller.

server/vehicle-persistence.lua
```diff
@@ -121,4 +132,50 @@ AddEventHandler('entityRemoved', function(entity)
...
+            }
+        end
+    end
+    return vehicles
+end)
+
+RegisterNetEvent('qbx_core:server:spawnVehicle', function(id, coords)
+    local vehicle = exports.qbx_vehicles:GetPlayerVehicle(id)
+    if not vehicle then return end
+    spawnVehicle(coords, id, vehicle.modelName, vehicle.props)
```
Manason, line 180: This is an exploitable event. Needs a server side check. Once a cache exists, you can use it to verify that the id vehicle exists in cache near the specified coords

Manason (changes requested):

server/vehicle-persistence.lua
```diff
@@ -1,3 +1,6 @@
+local persistence = GetConvarInt('qbx:enableVehiclePersistence', 0)
```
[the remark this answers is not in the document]
D4isDAVID: FYI, values `"true"` and `"false"` are interpreted as `1` and `0` respectively by `GetConvarInt`. But at this point this can actually be replaced by the new `GetConvarBool` I believe, which is available since artifact 10543.

server/vehicle-persistence.lua
```diff
@@ -121,4 +132,50 @@ AddEventHandler('entityRemoved', function(entity)
...
+
+local function spawnVehicle(coords, id, model, props)
+    local _, veh = qbx.spawnVehicle({
+        spawnSource = vector4(coords.x, coords.y, coords.z, coords.w),
+        model = model,
+        props = props
+    })
+    exports.qbx_core:EnablePersistence(veh)
+    Entity(veh).state:set('vehicleid', id, false)
+    SetVehicleDoorsLocked(veh, 2)
```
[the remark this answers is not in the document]
CodexisPhantom: Yeah I see, but there's no export in qbx_vehiclekeys to lock a car ?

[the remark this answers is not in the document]
Manason: You'll have to create one. Just extract the implementation of qb-vehiclekeys:server:setVehLockState into an export.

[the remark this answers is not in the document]
CodexisPhantom: Okey I see, I'll create a pr in qbx_vehiclekeys then.

server/vehicle-persistence.lua
```diff
@@ -121,4 +118,97 @@ AddEventHandler('entityRemoved', function(entity)
...
+    return cachedVehicles
+end)
+
+AddEventHandler('onResourceStart', function(resourceName)
+    if resourceName ~= 'qbx_vehicles' then return end
+
+    local vehicles = exports.qbx_vehicles:GetPlayerVehicles({ states = 0 })
+    if not vehicles then return end
+
+    for _, vehicle in ipairs(vehicles) do
```
BerkieBb, line 168: vehicles is an array, use a numeric for loop instead of this.

CodexisPhantom: All done

BerkieBb (approved):

CodexisPhantom:
Actually the position is saved only when there's a player in the car.

CodexisPhantom:
Not really I can implement that now

server/vehicle-persistence.lua
```diff
@@ -121,4 +118,135 @@ AddEventHandler('entityRemoved', function(entity)
...
+--- Save the vehicle position to the database
+---@param vehicle number
+---@param coords vector3
+---@param heading number
+local function saveVehiclePosition(vehicle, coords, heading)
+    local type = GetVehicleType(vehicle)
+    if type == 'heli' or type == 'plane' then
+        coords = vec3(coords.x, coords.y, coords.z + 1.0)
+    end
+    exports.qbx_vehicles:SaveVehicle(vehicle, {
```
BerkieBb, line 152: For style, add a space above this. Keep certain code blocks separate from each other.

server/vehicle-persistence.lua
```diff
@@ -121,4 +118,135 @@ AddEventHandler('entityRemoved', function(entity)
...
+AddEventHandler('onResourceStart', function(resourceName)
+    if resourceName ~= 'qbx_vehicles' then return end
+
+    local vehicles = exports.qbx_vehicles:GetPlayerVehicles({ states = 0 })
+    if not vehicles then return end
+
+    for i = 1, #vehicles do
+        local vehicle = vehicles[i]
+        if vehicle.coords and vehicle.props and vehicle.props.plate and
+            not isVehicleSpawned(vehicle.props.plate) then
```
BerkieBb, line 200: This being on the next line makes it look weird with the line below. Better to keep this on one line instead.

server/vehicle-persistence.lua
```diff
@@ -121,4 +118,135 @@ AddEventHandler('entityRemoved', function(entity)
...
+            cachedVehicles[vehicle.id] = vehicle.coords
+        end
+    end
+end)
+
+AddEventHandler('onResourceStop', function(resourceName)
+    if resourceName ~= cache.resource then return end
+    saveAllVehiclePosition()
```
BerkieBb, line 208: For style add a space above this.

server/vehicle-persistence.lua
```diff
@@ -121,4 +118,135 @@ AddEventHandler('entityRemoved', function(entity)
...
+end)
+
+AddEventHandler('txAdmin:events:scheduledRestart', function(eventData)
+    if eventData.secondsRemaining ~= 60 then return end
+    saveAllVehiclePosition()
```
BerkieBb, line 213: For style add a space above this.

PR #633, opened 2024-12-04
fix(client): ensure tutorial sessions

mafewtm:
## Description

The tutorial natives still seem to be causing issues as we're still getting reports of people unable to see other players. This edit forces the script to wait until the player is in or out of the session before it continues. I've tested this on local but not with other players.

I'm unsure why we're having such a hard time with these TutorialSession natives as ox_core uses them perfectly fine.

## Checklist

### DIFF
--- a/client/character.lua
+++ b/client/character.lua
@@ -102,6 +102,8 @@ local randomPeds = {
     }
 }
 
+NetworkStartSoloTutorialSession()
+
 local nationalities = {}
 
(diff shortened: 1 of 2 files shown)


Manason (approved):
We should test this or get someone else to test this in a multiplayer environment to make sure this fixes the issue or otherwise revert back to using buckets.

solareon (approved):

PR #634, opened 2024-12-04
fix(server/player): license login check passes for license and license2

Manason:
Fixes regression introduced by #618 which kicks players who have a license stored in database, since it doesn't check for db license matching license, only license2.

### DIFF
--- a/server/player.lua
+++ b/server/player.lua
@@ -38,8 +38,8 @@ function Login(source, citizenid, newData)
     end
 
     local license, license2 = GetPlayerIdentifierByType(source --[[@as string]], 'license'), GetPlayerIdentifierByType(source --[[@as string]], 'license2')
-    local userId = storage.fetchUserByIdentifier(license2 or license)
-
+    local userId = storage.fetchUserByIdentifier(license2) or storage.fetchUserByIdentifier(license)
+    assert(userId ~= nil, ('User does not exist. Licenses checked %s, %s'):format(license2, license))
     if citizenid then
         local playerData = storage.fetchPlayerEntity(citizenid)
         if playerData and (playerData.license == license2 or playerData.license == license) then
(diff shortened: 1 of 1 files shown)


server/player.lua
```diff
@@ -37,15 +37,14 @@ function Login(source, citizenid, newData)
         return false
     end
 
-    local license = GetPlayerIdentifierByType(source --[[@as string]], 'license2') or GetPlayerIdentifierByType(source --[[@as string]], 'license')
-    local userId = storage.fetchUserByIdentifier(license)
+    local license, license2 = GetPlayerIdentifierByType(source --[[@as string]], 'license'), GetPlayerIdentifierByType(source --[[@as string]], 'license2')
+    local userId = storage.fetchUserByIdentifier(license2 or license)
```
D4isDAVID, line 41: Should we check license and license2 here as well?

mafewtm (approved):

server/player.lua
Manason: @mafewtm What do you think?

mafewtm: whatevers better i guess

Manason: Attempting to fetch by both now and also asserting that a user entity exists in the database.

server/player.lua
```diff
@@ -37,15 +37,14 @@ function Login(source, citizenid, newData)
         return false
     end
 
-    local license = GetPlayerIdentifierByType(source --[[@as string]], 'license2') or GetPlayerIdentifierByType(source --[[@as string]], 'license')
-    local userId = storage.fetchUserByIdentifier(license)
-
+    local license, license2 = GetPlayerIdentifierByType(source --[[@as string]], 'license'), GetPlayerIdentifierByType(source --[[@as string]], 'license2')
+    local userId = storage.fetchUserByIdentifier(license2) or storage.fetchUserByIdentifier(license)
+    assert(userId ~= nil, ('User does not exist. Licenses checked %s, %s'):format(license2, license))
```
D4isDAVID, line 42: I think we should use `lib.print.error` and return false rather than throwing an error, similarly to what we do when validating the `source` variable.

D4isDAVID (changes requested):

server/player.lua
Manason: done

D4isDAVID (approved):

PR #645, opened 2024-12-22
feat(modules): add duffel bag check

5hahfivem:
## Description

This pull request adds a feature to check if a player is wearing a duffel bag. 

## Checklist

### DIFF
--- a/modules/lib.lua
+++ b/modules/lib.lua
@@ -89,6 +89,37 @@ qbx.armsWithoutGloves = lib.table.freeze({
     }),
 })
 
+
+qbx.duffelbagIndexes = lib.table.freeze({
+    male = lib.table.freeze({
+        [40] = true,
+        [41] = true,
+        [44] = true,
+        [45] = true,
+        [81] = true,
+        [82] = true,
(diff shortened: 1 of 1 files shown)


Manason (commented):
What's the use cases for this? It seems pretty specific to be in lib. Are we going to end up with 100s of functions to check if they are wearing x model?

EchoStudioss:
Currently have it running on my bank robbery, back items. Anything that would realistically require a duffel bag, I did it for myself, but I PR'ed it since the iswearinggloves ft is handy, with the use cases for a duffel bag requirement I thought it to be equally as useful for others.

FjamZoo:
I don't think this belongs in a "core", it just ends up having a lot of bloated functions.

Even the "glove" checks dosn't belong in the core and should be moved to somewhere like Evidence.

Manason (approved):
I'm still a bit on the fence about whether this makes sense, but it is in the same category as gloves, so unless we can think of a better place, it could be useful to many scripts to have this and therefore it makes sense to be in core.

PR #649, opened 2024-12-26
fix(server/money): setMoney

Frowmza:
Figured this out after receiving reports that my banking script wasn’t working on QBX correctly.

## Description
- Corrected the difference calculation for setMoney. The previous state always returned 0 because the calculation was performed after setting the player's money.
- Pass the amount instead of the difference to setMoney. This aligns with QB, as many scripts, including my banking script, were designed to handle the new amount directly.
- For `RemoveMoney` and `AddMoney`, the difference was incorrectly passed as false for both in `hud:client:OnMoneyChange`. I’m not sure if this is being used, but it should be true for remove and false for add as in `QB`. Used a ternary operator to handle this logic appropriately

## Checklist

### DIFF
--- a/server/player.lua
+++ b/server/player.lua
@@ -1195,7 +1195,7 @@ local function emitMoneyEvents(source, playerMoney, moneyType, amount, actionTyp
     local isSet = actionType == 'set'
     local isRemove = actionType == 'remove'
 
-    TriggerClientEvent('hud:client:OnMoneyChange', source, moneyType, isSet and difference or amount, isSet and difference < 0 or isRemove and true or false, reason)
+    TriggerClientEvent('hud:client:OnMoneyChange', source, moneyType, isSet and difference and math.abs(difference) or amount, isSet and difference and difference < 0 or isRemove and true or false, reason)
     TriggerClientEvent('QBCore:Client:OnMoneyChange', source, moneyType, amount, actionType, reason)
     TriggerEvent('QBCore:Server:OnMoneyChange', source, moneyType, amount, actionType, reason)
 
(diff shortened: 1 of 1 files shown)


mafewtm (approved):

server/player.lua
```diff
@@ -1189,14 +1189,17 @@ exports('SetCharInfo', SetCharInfo)
...
----@param direction boolean
+---@param difference number
 ---@param reason? string
-local function emitMoneyEvents(source, playerMoney, moneyType, amount, actionType, direction, reason)
-    TriggerClientEvent('hud:client:OnMoneyChange', source, moneyType, amount, direction)
+local function emitMoneyEvents(source, playerMoney, moneyType, amount, actionType, difference, reason)
+    local isSet = actionType == 'set'
+    local isRemove = actionType == 'remove'
+
+    TriggerClientEvent('hud:client:OnMoneyChange', source, moneyType, isSet and difference or amount, isSet and difference < 0 or isRemove and true or false, reason)
```
Manason, line 1198:
```suggestion
    TriggerClientEvent('hud:client:OnMoneyChange', source, moneyType, isSet and difference or amount, isSet and difference < 0 or isRemove, reason)
```

Manason: difference will never be less than 0 since we're passing in the absolute value. No behavior change is desired compared to qb-core's spec for this event https://github.com/qbcore-framework/qb-core/blob/main/server/player.lua#L377

Manason (changes requested):

server/player.lua
Frowmza: ok i passed raw the difference then we switch it to absolute value

Frowmza: also, should i make the difference param the latest, and just pass it as nil for `RemoveMoney` and `AddMoney`

Manason: Not sure what you mean by that

Manason (approved):

server/player.lua
Frowmza: i did change `difference` param order, and make it the last

Manason: Oh, I see. Yeah that's fine.

PR #651, opened 2024-12-29
fix(lib/spawn): retrieve vehicle type

Frowmza:
## Description

- use callback on random client instead of server spawn.
`CreateVehicle` on the server side still depends on the client to spawn. If no client is nearby, it will get stuck in a while loop and vehicle will never exist. Instead, we retrieve the type from a random client, which will definitely exist.

## Checklist

### DIFF
--- a/modules/lib.lua
+++ b/modules/lib.lua
@@ -339,6 +339,18 @@ if isServer then
         return netId, veh
     end
 else
+    ---@param model number|string
+    ---@return string
+    lib.callback.register('qbx_core:client:getVehicleType', function(model)
+        model = lib.requestModel(model)
+        local coords = GetEntityCoords(cache.ped)
+        local tempVehicle = CreateVehicle(model, coords.x, coords.y, coords.z - 20.0, 0, false, false)
+        SetModelAsNoLongerNeeded(model)
+        local type = GetVehicleType(tempVehicle)
+        DeleteEntity(tempVehicle)
+        return type
+    end)
+
     ---@class LibDrawTextParams
     ---@field text string
     ---@field scale? integer default: `0.35`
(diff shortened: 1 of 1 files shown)


Manason (commented):
1. Why does it matter if creating a vehicle is stuck waiting for a client? It will be anyway as CreateVehicleServerSetter blocks waiting for a nearby client.

2. In my experience the current way of getting vehicle types is very reliable. What happens if there is no client connected?

Frowmza:
> 1. Why does it matter if creating a vehicle is stuck waiting for a client? It will be anyway as CreateVehicleServerSetter blocks waiting for a nearby client.
I don't get what you mean here 

> 2\. What happens if there is no client connected?
i can add check if `GetPlayers` is empty.

I tested this an hour ago. The vehicle I was trying to spawn does not exist in `GetVehiclesByHash`. Instead, it attempts to create the vehicle, but it gets stuck at `while not DoesEntityExist() do`.

Manason:
> > 1. Why does it matter if creating a vehicle is stuck waiting for a client? It will be anyway as CreateVehicleServerSetter blocks waiting for a nearby client.
> 
> I don't get what you mean here
> 
> > 2. What happens if there is no client connected?
> 
> i can add check if `GetPlayers` is empty.
> 
> I tested this an hour ago. The vehicle I was trying to spawn does not exist in `GetVehiclesByHash`. Instead, to get the type, it attempts to create the vehicle, but it gets stuck at `while not DoesEntityExist() do`.

Does it get permanently stuck, or does it resolve when you get close?

Frowmza:
Just retested this now, it seems the issue was on my end. I was spawning vehicles directly on resource start, which doesn’t work. After adding a wait(100) before spawning, it seems to be working as intended. Sorry for the trouble!

PR #653, opened 2025-01-04
feat(server/player): support getting/setting nested metadata

### LINKED ISSUE #646: Set/Get Metadata nested key handling
### The problem

exports.qbx_core:SetMetadata(id, 'licences.driver', true) results in

```lua
['licences.driver'] = true
```
rather than
```lua
licences = {
  driver = true
}
```

### Ideal solution

Handle nested table entries so that one can set and get metadata at a specific nested key

### Alternative solutions

Do nothing and continue to force getting and setting by top level key only

### Additional context

_No response_

mafewtm:
## Description

Handles nested metadata and closes #646. Doesn't allow for more than one level deep. I think there's probably a better way to do this but right now it just checks for periods as seperators.

## Checklist

### DIFF
--- a/server/player.lua
+++ b/server/player.lua
@@ -1124,9 +1124,23 @@ function SetMetadata(identifier, metadata, value)
 
     if not player then return end
 
-    local oldValue = player.PlayerData.metadata[metadata]
+    local oldValue
 
-    player.PlayerData.metadata[metadata] = value
+    if metadata:match('%.') then
+        local metaTable, metaKey = metadata:match('([^%.]+)%.(.+)')
+
+        if metaKey:match('%.') then
+            lib.print.error('cannot get nested metadata more than 1 level deep')
+        end
+
+        oldValue = player.PlayerData.metadata[metaTable][metaKey]
+
+        player.PlayerData.metadata[metaTable][metaKey] = value
+    else
+        oldValue = player.PlayerData.metadata[metadata]
+
+        player.PlayerData.metadata[metadata] = value
+    end
 
     UpdatePlayerData(identifier)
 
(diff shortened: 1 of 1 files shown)


Manason (commented):
What if we split the key on periods using gmatch, then iterate over the nested keys to have this solution work for any depth. One thing I didn't think of, could this be a problem for existing metadata event handlers. i.e. if someone is listening for a change in the key 'license', then this solution would also need to emit the event for the top level key in addition to the nested key string to preserve backwards compatibility.

mafewtm:
> What if we split the key on periods using gmatch, then iterate over the nested keys to have this solution work for any depth. One thing I didn't think of, could this be a problem for existing metadata event handlers. i.e. if someone is listening for a change in the key 'license', then this solution would also need to emit the event for the top level key in addition to the nested key string to preserve backwards compatibility.

I believe it will be an issue for those people. No good way to work around that though.

ChatDisabled:
Tested with
```lua
AddEventHandler('qbx_core:server:onSetMetaData', function(metadata,  oldValue, value, playerSource)
    print(metadata)
    print(json.encode(oldValue, {indent = true}))
    print(json.encode(value, {indent = true}))
end)

exports.qbx_core:SetMetadata(1, 'licences.driver', false)
print(exports.qbx_core:GetMetadata(1, 'licences.driver'))
print('--')
exports.qbx_core:SetMetadata(1, 'licences.driver', true)
print(exports.qbx_core:GetMetadata(1, 'licences.driver'))
```
Which results in
```
[script:testing] licences
[script:testing] {
[script:testing]     "weapon": false,
[script:testing]     "driver": false,
[script:testing]     "id": true
[script:testing] }
[script:testing] false
[script:testing] false
[script:testing] --
[script:testing] licences
[script:testing] {
[script:testing]     "weapon": false,
[script:testing]     "driver": true,
[script:testing]     "id": true
[script:testing] }
[script:testing] true
[script:testing] true
```

So this still allows for doing `exports.qbx_core:SetMetadata(id, 'licences.driver', true)` and `exports.qbx_core:GetMetadata(1, 'licences.driver')` while maintaining the old behaviour in case some people listen to the top level key.

Feel free to update this to allow for even deeper nested keys but we don't really even have that to begin with besides `apartmentType` and `apartmentId` (these aren't even used by us they are old QB remnants so we never use a metadata deeper than 1). I wouldn't really give people this option to begin with since the whole MetaData is a big mess of ugly json in SQL to begin with.

PR #655, opened 2025-01-09
feat(modules/lib): deprecate qbx.playAudio

### LINKED ISSUE #492: [SUGGESTION] Sound Broadcast
### The problem

The core has a playAudio function. This is a useful tool, but it has a limitation. The played sound is only heard by one player.

### Ideal solution

The function has a brodcast parameter so that everyone can hear the sound.

### Alternative solutions

_No response_

### Additional context

The parameter can be useful, for example, when unlocking a vehicle.

mafewtm:
## Description

Marks `qbx.playAudio` as deprecated since we now utilize `mana_audio`. Closes #492 upon merging.

## Checklist

### DIFF
--- a/modules/lib.lua
+++ b/modules/lib.lua
@@ -578,6 +578,7 @@ else
     ---@field audioSource? number | vector3 entity handle or vector3 coords
     ---@field range? number only used if `audioSource` is a vector3 coordinate
 
+    ---@deprecated use mana_audio instead
     ---Plays a sound with the provided audio name and audio ref.
     ---If `returnSoundId` is false or not specified the soundId is released,
     ---otherwise the function returns the soundId without releasing it.


Manason (commented):
mana_audio doesn't return the soundId so not sure it's exactly a 1:1 replacement

mafewtm:
> mana_audio doesn't return the soundId so not sure it's exactly a 1:1 replacement

It's not meant to be a one of one replacement. The issue was brought up about it being unnecessary, pointless, not networked, and now that we have our own sound resource that relatively the same thing, but better, I decided to toss a PR in for deprecation. If we wanna keep it then that's fine. I just figured we wanna steer people to use mana_audio instead

Manason (approved):
I guess we can always improve mana_audio to handle more than just sound clips if we want.

PR #656, opened 2025-01-09
chore(server/events): validation creating user

nurrifqialhusaini:
## Description

Balidate and prevent false create user when someone trying "abuse" login

### DIFF
--- a/server/events.lua
+++ b/server/events.lua
@@ -101,7 +101,9 @@ local function onPlayerConnecting(name, _, deferrals)
 
         identifiers.username = name
 
-        storage.createUser(identifiers)
+        if license then
+            storage.createUser(identifiers)
+        end
     end
 
     local databaseTime = os.clock()


nurrifqialhusaini:
Balidate hahaha, got typo for desc

solareon:
This would already be caught by https://github.com/nurrifqialhusaini/qbx_core/blob/1fc0d379b57949ca07bfaa7ca0cff34ddd546968/server/events.lua#L93-L97 also additionally this code is only run server side from an event triggered by the server so not entirely sure how this could be abused. Do you have a working proof of concept for an exploit of this?

nurrifqialhusaini:
for proff its difficult, maybe bcs im rework the license to 'steam' or 'fivem'. and if this not get it still can joining the server, whit null data. Of course this method already modified the base structure. but prevention is better than cure right.. makes it safer when a dev will customize without worrying about abuse.

BerkieBb:
You're writing duplicate code here, the license already gets checked a few lines before your change. It stops you from entering if your license is not defined.

PR #659, opened 2025-01-11
Update events.lua

SKITTLE6969:
This change intend to stop the player count from going above the set amount in places like rich presence

## Checklist


solareon (approved):
I presume there is a minimum citizenfx version that this got added in? Also is there anywhere else that we modify this state?

SKITTLE6969:
> I presume there is a minimum citizenfx version that this got added in? Also is there anywhere else that we modify this state?

I've been having this on my version since last Feb so I'm pretty sure citizenfx version won't cause any issues with anything after that before that I haven't tested.
So far this is the only instance I've seen this change required but will do a thorough search tomorrow and do the needful

PR #661, opened 2025-01-14
feat: live database updates

BerkieBb:
## Description

This PR introduces Live Database Updates. It changes how player data is updated by making sure only the any updated playerdata field is updated specifically in the database each 15 seconds (configurable). This is not directly live, but live enough to ensure data gets updated fast.

I also cleaned up the codebase a bit with this PR.

I have tested this myself and fixed bugs along but it's best to make sure this gets tested with others as well. However don't test in production until a few have tested in a test environment.

This also adds 2 new events that will be triggered from SetPlayerData in qbx_core/server/player.lua
qbx_core:client:setPlayerData
qbx_core:server:setPlayerData

## Checklist

### DIFF
--- a/server/player.lua
+++ b/server/player.lua
@@ -138,7 +138,7 @@ function SetJobDuty(identifier, onDuty)
     local player = type(identifier) == 'string' and (GetPlayerByCitizenId(identifier) or GetOfflinePlayer(identifier)) or GetPlayer(identifier)
 
     if not player then
-        error(('SetJobDuty couldn\'t find player with identifier %s'):format(identifier))
+        lib.print.error(('SetJobDuty couldn\'t find player with identifier %s'):format(identifier))
         return
     end
 
@@ -1078,7 +1078,7 @@ function SetPlayerData(identifier, key, value, cb, cancelDbUpdate)
     local player = type(identifier) == 'string' and (GetPlayerByCitizenId(identifier) or GetOfflinePlayer(identifier)) or GetPlayer(identifier)
 
     if not player then
-        error(('SetPlayerData couldn\'t find player with identifier %s'):format(identifier))
+        lib.print.error(('SetPlayerData couldn\'t find player with identifier %s'):format(identifier))
         return
     end
 
@@ -1094,7 +1094,7 @@ function SetPlayerData(identifier, key, value, cb, cancelDbUpdate)
             elseif i ~= (#key - 1) then
                 -- if an invalid key is specified and we are not on the last one, stop trying to update
                 -- reason for allowing the last one to not exist is so we can insert new values
-                error(('key %s doesn\'t exist in table player.PlayerData.%s'):format(key[i], key[1]))
+                lib.print.error(('key %s doesn\'t exist in table player.PlayerData.%s'):format(key[i], key[1]))
                 return
             end
         end
--- a/server/player.lua
+++ b/server/player.lua
@@ -135,6 +135,13 @@ exports('SetJob', SetJob)
 ---@param identifier Source | string
 ---@param onDuty boolean
 function SetJobDuty(identifier, onDuty)
+    local player = type(identifier) == 'string' and (GetPlayerByCitizenId(identifier) or GetOfflinePlayer(identifier)) or GetPlayer(identifier)
+
+    if not player then
+        error(('SetJobDuty couldn\'t find player with identifier %s'):format(identifier))
+        return
+    end
+
     SetPlayerData(identifier, 'job', {'onduty'}, not not onDuty, function()
         TriggerEvent('QBCore:Server:SetDuty', player.PlayerData.source, player.PlayerData.job.onduty)
         TriggerClientEvent('QBCore:Client:SetDuty', player.PlayerData.source, player.PlayerData.job.onduty)
@@ -196,7 +203,7 @@ function SetPlayerPrimaryJob(citizenid, jobName)
 
     player.PlayerData.job = toPlayerJob(jobName, job, grade)
 
-    SetPlayerData(player.PlayerData.source, 'job', nil, player.PlayerData.job, function()
+    SetPlayerData(citizenid, 'job', nil, player.PlayerData.job, function()
         TriggerEvent('QBCore:Server:OnJobUpdate', player.PlayerData.source, player.PlayerData.job)
         TriggerClientEvent('QBCore:Client:OnJobUpdate', player.PlayerData.source, player.PlayerData.job)
     end)
@@ -261,7 +268,7 @@ function AddPlayerToJob(citizenid, jobName, grade)
     storage.addPlayerToJob(citizenid, jobName, grade)
 
     player.PlayerData.jobs[jobName] = grade
-    SetPlayerData(player.PlayerData.source, 'jobs', nil, player.PlayerData.jobs, function()
+    SetPlayerData(citizenid, 'jobs', nil, player.PlayerData.jobs, function()
         TriggerEvent('qbx_core:server:onGroupUpdate', player.PlayerData.source, jobName, grade)
         TriggerClientEvent('qbx_core:client:onGroupUpdate', player.PlayerData.source, jobName, grade)
     end, true)
@@ -307,10 +314,10 @@ function RemovePlayerFromJob(citizenid, jobName)
         local job = GetJob('unemployed')
         assert(job ~= nil, 'cannot find unemployed job. Does it exist in shared/jobs.lua?')
 
-        SetPlayerData(player.PlayerData.source, 'job', nil, toPlayerJob('unemployed', job, 0))
+        SetPlayerData(citizenid, 'job', nil, toPlayerJob('unemployed', job, 0))
     end
 
-    SetPlayerData(player.PlayerData.source, 'jobs', nil, player.PlayerData.jobs, function()
+    SetPlayerData(citizenid, 'jobs', nil, player.PlayerData.jobs, function()
         TriggerEvent('qbx_core:server:onGroupUpdate', player.PlayerData.source, jobName)
         TriggerClientEvent('qbx_core:client:onGroupUpdate', player.PlayerData.source, jobName)
     end, true)
@@ -396,7 +403,7 @@ function SetPlayerPrimaryGang(citizenid, gangName)
 
     assert(gang.grades[grade] ~= nil, ('gang %s does not have grade %s'):format(gangName, grade))
 
-    SetPlayerData(player.PlayerData.source, 'gang', nil, {
+    SetPlayerData(citizenid, 'gang', nil, {
         name = gangName,
         label = gang.label,
         isboss = gang.grades[grade].isboss,
@@ -468,7 +475,7 @@ function AddPlayerToGang(citizenid, gangName, grade)
     storage.addPlayerToGang(citizenid, gangName, grade)
 
     player.PlayerData.gangs[gangName] = grade
-    SetPlayerData(player.PlayerData.source, 'gangs', nil, player.PlayerData.gangs, function()
+    SetPlayerData(citienid, 'gangs', nil, player.PlayerData.gangs, function()
         TriggerEvent('qbx_core:server:onGroupUpdate', player.PlayerData.source, gangName, grade)
         TriggerClientEvent('qbx_core:client:onGroupUpdate', player.PlayerData.source, gangName, grade)
     end, true)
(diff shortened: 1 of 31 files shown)


server/storage/players.lua
```diff
@@ -382,8 +418,99 @@ local function cleanPlayerGroups()
...
+                                updateStrings = { ('$.%s.%s'):format(k, k2), v2 }
+                            end
+                        end
+                    else
+                        updateStrings[#updateStrings + 1] = { ('$.%s'):format(k), v }
+                    end
+                end
+
+                for i = 1, #updateStrings do
+                    MySQL.prepare.await(('UPDATE players SET %s = JSON_SET(%s, "%s", ?) WHERE citizenid = ?'):format(key, key, updateStrings[i][1]), { updateStrings[i][2], citizenid })
```
D4isDAVID, line 498: Is it better to loop and make multiple requests like this, or maybe to create a bulk request will all the changes at once? Or maybe some sort of combination of both?

server/storage/players.lua
```diff
@@ -382,8 +418,99 @@ local function cleanPlayerGroups()
...
+                for k, v in pairs(data) do
+                    if type(v) == 'table' then
+                        for k2, v2 in pairs(v) do
+                            if type(v2) == 'table' then
+                                for k3, v3 in pairs(v2) do
+                                    if type(v3) == 'table' then
+                                        v3 = json.encode(v3)
+                                    end
+
+                                    updateStrings[#updateStrings + 1] = { ('$.%s.%s.%s'):format(k, k2, k3), v3 }
```
D4isDAVID, line 486: Not sure how I feel about this part. Shouldn't we create this type of table in `addPlayerDataUpdate` with a string of the keys and the value, instead of creating a nested table through loops and then walking through the tables in this function?

server/player.lua
```diff
@@ -135,16 +137,15 @@ exports('SetJob', SetJob)
...
-    if not player then return end
-
-    player.PlayerData.job.onduty = not not onDuty
-
-    if player.Offline then return end
-
-    TriggerEvent('QBCore:Server:SetDuty', player.PlayerData.source, player.PlayerData.job.onduty)
-    TriggerClientEvent('QBCore:Client:SetDuty', player.PlayerData.source, player.PlayerData.job.onduty)
+    if not player then
+        error(('SetJobDuty couldn\'t find player with identifier %s'):format(identifier))
```
D4isDAVID, line 141:
```suggestion
        lib.print.error(('SetJobDuty couldn\'t find player with identifier %s'):format(identifier))
```
This is a breaking change as `error` will cause the parent code to stop running without a `pcall`. Let's change it to `lib.print.error` which doesn't error out.

server/player.lua
```diff
@@ -1081,29 +1068,62 @@ exports('SaveOffline', SaveOffline)
 
 ---@param identifier Source | string
 ---@param key string
+---@param subKeys? string[]
 ---@param value any
-function SetPlayerData(identifier, key, value)
+---@param cb? function A function that's called after the standard SetPlayerData events are triggered if the player is online
+---@param cancelDbUpdate? boolean When true, makes sure the database doesn't get updated as a result of this change
+function SetPlayerData(identifier, key, subKeys, value, cb, cancelDbUpdate)
```
D4isDAVID, line 1075: How about we accept a `key` parameter of both type string and a table of strings instead of introducing the `subKeys` parameter?

D4isDAVID (changes requested):

server/player.lua
BerkieBb: You're right, I forgot error stops execution

server/player.lua
BerkieBb: Sounds better indeed, I'll apply that

server/storage/players.lua
BerkieBb: I do it like this so I can simulate partial playerdata updates basically when adding and when sending only converting it to queries. If I don't do it this way and immediately make these strings, we can have duplicate strings having a different value which can possibly cause an issue.

server/storage/players.lua
BerkieBb: The idea of a transaction slipped my mind. Prepared statements just seemed the best since it will update frequently, but it's not very performant against a transaction possibly.

server/player.lua
```diff
@@ -1073,37 +1059,71 @@ function SaveOffline(playerData)
...
----@param key string
+---@param key string | string[]
 ---@param value any
-function SetPlayerData(identifier, key, value)
-    if type(key) ~= 'string' then return end
+---@param cb? function A function that's called after the standard SetPlayerData events are triggered if the player is online
+---@param cancelDbUpdate? boolean When true, makes sure the database doesn't get updated as a result of this change
+function SetPlayerData(identifier, key, value, cb, cancelDbUpdate)
+    local hasSubKeys = type(key) == 'table'
+    if type(key) ~= 'string' or not hasSubKeys then return end
```
Frowmza, line 1076: shouldn't be `and`?

server/player.lua
```diff
@@ -1073,37 +1059,71 @@ function SaveOffline(playerData)
...
+        return
+    end
 
-    player.PlayerData[key] = value
+    local oldValue = player.PlayerData[hasSubKeys and key[1] or key]
+
+    if hasSubKeys then
+        local current = player.PlayerData[hasSubKeys and key[1] or key]
+        -- We don't check the last one because otherwise we lose the table reference
+        for i = 2, #key - 1 do
```
Frowmza, line 1090: Here, if we have nested player data like `{'job', 'onduty'}` for `key`, then `#key - 1` would equal 1, i think loop wouldn't work because 2 is greater than 1, it like doing `for i = 2, 1 do`, ill assume it should be starting from 1 and not 2

server/storage/players.lua
```diff
@@ -382,8 +418,98 @@ local function cleanPlayerGroups()
...
+    if not currentTable[citizenid] then
+        currentTable[citizenid] = {}
+    end
+
+    currentTable[citizenid][hasSubKeys and key[1] or key] = hasSubKeys and {} or value
+
+    if hasSubKeys then
+        local current = currentTable[citizenid][key[1]]
+        -- We don't check the last one because otherwise we lose the table reference
+        for i = 2, #key - 1 do
```
Frowmza, line 461: same for this

server/storage/players.lua
```diff
@@ -382,8 +418,98 @@ local function cleanPlayerGroups()
...
+                        end
+                    else
+                        MySQL.prepare.await(('UPDATE players SET %s = JSON_SET(%s, "$.%s", ?) WHERE citizenid = ?'):format(key, key, k), { v, citizenid })
+                    end
+                end
+            else
+                MySQL.prepare.await(('UPDATE players SET %s = ? WHERE citizenid = ?'):format(key), { data, citizenid })
+            end
+        end
+    end
```
Frowmza, line 503:
```suggestion
    ---@type function
    local prepare_await = MySQL.prepare.await

    ---comment
    ---@param key string
    ---@param nestedTable table<string, any>
    ---@param path string?
    ---@param citizenid string
    local function updateNestedPlayerData(key, nestedTable, citizenid, path)
        for k, v in pairs(nestedTable) do
            local currentPath = path and ('%s.%s'):format(path, k) or k

            if type(v) == 'table' then
                updateNestedPlayerData(key, v, citizenid, currentPath)
            else
                local query = ('UPDATE players SET %s = JSON_SET(%s, "$.%s", ?) WHERE citizenid = ?'):format(key, key, currentPath)
                prepare_await(query, { v, citizenid })
            end
        end
    end

    for citizenid, playerData in pairs(collectedPlayerData) do
        for key, data in pairs(playerData) do
            if type(data) == 'table' then
                updateNestedPlayerData(key, data, citizenid)
            else
                local query = ('UPDATE players SET %s = ? WHERE citizenid = ?'):format(key)
                prepare_await(query, { data, citizenid })
            end
        end
    end
```

Here i tried to make this readable and easy to read, as i used same trick in my bl_bridge

Frowmza (changes requested):

server/player.lua
BerkieBb: Nope, it has to be either one of them, not both at the same time

server/player.lua
BerkieBb: It should be starting from 2 since `current` is the one that implements the first key already. But I did overlook it having just 2 keys with that one. I'll fix it tomorrow. Reason for it doing -1 is already written there

server/player.lua
Frowmza: But you are returning

BerkieBb: Ah shit I now realize the issue lol

server/storage/players.lua
BerkieBb: This is a good one, but a slight part is wrong and that is that you make the path start with the key. it can't do that because the key is the column it is in, so you can't start with that because the dollar sign in the query already refers to that. I'll adjust it though so it is good.

BerkieBb: It does add the risk of someone making their playerdata so big that they nest tables so much that it can crash the server, but eh, it's on them then.

BerkieBb: I wanted to avoid that, but the cost of readability is not worth it I see as you need like 5 or more nests to crash

server/storage/players.lua
```diff
@@ -382,8 +418,97 @@ local function cleanPlayerGroups()
...
+    currentTable[citizenid][hasSubKeys and key[1] or key] = hasSubKeys and {} or value
+
+    if hasSubKeys then
+        local current = currentTable[citizenid][key[1]]
+        if #key > 2 then
+            -- We don't check the last one because otherwise we lose the table reference
+            for i = 2, #key - 1 do
+                if not current[key[i]] then
+                    current[key[i]] = {}
+                end
```
Frowmza, line 465:
```suggestion
                    current[key[i]] = {}
                    current = current[key[i]]
                end
```
Also this, we need to update the `current` so we can index the last nested key and insert the value

server/storage/players.lua
```diff
@@ -382,8 +418,97 @@ local function cleanPlayerGroups()
...
+        lib.print.error(('Tried to update player data field %s when it doesn\'t exist. Value: %s'):format(hasSubKeys and key[1] or key, value))
+        return
+    end
+
+    if hasSubKeys and not jsonPlayerFields[key[1]] then
+        lib.print.error(('Tried to update player data field %s as a json object when it isn\'t one'):format(key[1]))
+        return
+    end
+
+    value = type(value) == 'table' and json.encode(value) or value
```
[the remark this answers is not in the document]
BerkieBb: When it's encoded, it becomes a string, so no, it would not encode an encoded table.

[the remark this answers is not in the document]
BerkieBb: We don't encode anywhere else besides that line, so I'm not sure I understand your point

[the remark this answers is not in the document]
Frowmza: So if value is a `table` it will encode it, also it will encode and set the encoded value in the query, because we using `JSON_SET()`, i'm not great at explaining but i did tested it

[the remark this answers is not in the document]
Frowmza: you can try and test it here the command
```lua
RegisterCommand('setplayerdata', function(source, args)
    SetPlayerData('P89Q0713', {'job', 'grade'}, {level = 0, name = 'test2'})
    storage.sendPlayerDataUpdates()
end)
```
make sure this cid is exist in players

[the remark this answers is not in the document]
Frowmza: image
this how it will look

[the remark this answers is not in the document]
BerkieBb: It's not encoding twice because of the query. We already send a string to it so it just handles it like a string and for json, that works.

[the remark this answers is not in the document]
BerkieBb: Oh nevermind then, that's odd. And without encoding it does it correctly?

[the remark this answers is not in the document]
Frowmza: yeah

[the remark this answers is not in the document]
BerkieBb: Can do a suggestion to remove it, then I can commit it from my phone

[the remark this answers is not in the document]
Frowmza:
```suggestion

```

Frowmza (changes requested):

BerkieBb:
Ah damnit now it created a 2-space gap in between

Frowmza (approved):

ChatDisabled:
Look I get it. But I have to agree with the team that this is just aids to read through. Pretty please open separate PRs, preferably one big `style:` first so we can immediately merge it. And then from there apply your changes. Trying to tell server owners to test this out is hard when they have to scroll 10 pages before finding a relevant line change.

PR #663, opened 2025-01-24
Optimization: Update Player Data When Key Change

sleepyexe:
## Description
This change is to update player data when the key of data changing, so we can reduce the traffic from server to client.
Also i tested it for qb-core compat.

## Checklist

### DIFF
--- a/client/events.lua
+++ b/client/events.lua
@@ -20,6 +20,18 @@ RegisterNetEvent('QBCore:Client:OnPlayerLoaded', function()
     end
 end)
 
+RegisterNetEvent('qbx_core:client:updatePlayerData', function (key, value)
+    if source == '' then return end
+    QBX.PlayerData[key] = value
+    TriggerEvent('QBCore:Player:SetPlayerData', QBX.PlayerData)
+end)
+
+RegisterNetEvent('qbx_core:client:loadPlayerData', function (val)
+    if source == '' then return end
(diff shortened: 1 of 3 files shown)


KrisnaCahya (approved):

nurrifqialhusaini (approved):
nice, optimized and better trafic

xtrsyz (approved):

BerkieBb:
My PR #661 will change a lot of how player data is saved and updated inside the core. You should wait until that one is merged and adjust this one.

jellyton255:
The blocker for this PR was closed, this can probably be something that can be considered for merging now.

jellyton255:
I've implemented these changes on my server, and the PR itself works fine, however, some external resources use `if source == '' then return end` in their resources in the player data event handler, which now blocks the `QBCore:Player:SetPlayerData` event.

sleepyexe:
> I've implemented these changes on my server, and the PR itself works fine, however, some external resources use `if source == '' then return end` in their resources in the player data event handler, which now blocks the `QBCore:Player:SetPlayerData` event.

Thanks for testing it out! The issue seems to stem from external resources using if source == '' then return end, which now blocks QBCore:Player:SetPlayerData. If those resources are not escrowed, they should be able to adjust their validation to accommodate this change.

PR #665, opened 2025-01-27
Update character.lua

WhereiamL:
Fixed a Black Screen Issue: Resolved an issue where players encountered a black screen when creating an identity with "starting apartments" enabled.

## Description

This pull request introduces the following improvements:

Bug Fix: Added a fallback mechanism to handle cases where required resources (qbx_spawn or qbx_apartments) are missing. If a resource is unavailable, the system defaults to spawnDefault, preventing the black screen issue.

Code Refactor: Encapsulated the spawn logic into a modular and reusable function. This improves readability, maintainability, and makes the codebase easier to extend in the future.

Inline Documentation: Added LuaDoc comments to describe function parameters, return types, and key logic flows. This helps other developers quickly understand the purpose and usage of the code.

## Checklist

### DIFF
--- a/client/character.lua
+++ b/client/character.lua
@@ -116,6 +116,12 @@ if config.characters.limitNationalities then
     end)
 end
 
+--- @param resourceName string The name of the resource to check.
+--- @return boolean True if the resource is missing, false otherwise.
+local function isResourceMissing(resourceName)
+    return GetResourceState(resourceName) == 'missing'
+end
+
 local function setupPreviewCam()
     DoScreenFadeIn(1000)
     SetTimecycleModifier('hud_def_blur')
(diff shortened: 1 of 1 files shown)


client/character.lua
```diff
@@ -116,6 +116,12 @@ if config.characters.limitNationalities then
     end)
 end
 
+--- @param resourceName string The name of the resource to check.
+--- @return boolean True if the resource is missing, false otherwise.
+local function isResourceMissing(resourceName)
+    return GetResourceState(resourceName) == 'missing'
```
Frowmza, line 122: what if resource is `stopped` , `uninitialized` or any other state, would be better just `~= 'started'`

Frowmza (changes requested):

client/character.lua
mafewtm: There's also no real need to make that a function

WhereiamL: > what if resource is `stopped` , `uninitialized` or any other state, would be better just `~= 'started'`

Even better I did not check that qbx_apartments is achieved

WhereiamL: > There's also no real need to make that a function

It's user friendly for people who don't know a lot, and as u can see on the discord there are multiple people with the black screen issue

mafewtm: `isResourceMissing(qbx_spawn)` says the exact same thing as `GetResourceState('qbx_spawn') == 'missing'`. It's unneeded

Frowmza: I don't mind, but I think since we're already using GetResourceState with its method in the same file, we should just continue using it so `GetResourceState('qbx_spawn'):find('start')` would be better.

WhereiamL (commented):
Changes

Issue #235, opened 2023-11-12
Change salary as a Boss/Owner of a job


Birdy113:
### The problem

You cannot change salary as boss/owner.  From old days when old ESX we could do that but not with newer frameworks. I miss this feauture 

### Ideal solution

Give freedom to business owners (Boss) to change salary to their employees to create better business RP

### Alternative solutions

Make it in DB salary system or make it possible to do salary in DB. (Changeable)

Manason:
Yeah this is something I'd love to do in the future when I have some time. The way we'd do it is adding a separate jobs table which lists all the job names and their salaries. Then the existing jobs config would just be the default salary if there is no value in the database.

If someone can get to it before me though that would be great.

TonybynMp4:
> Yeah this is something I'd love to do in the future when I have some time. The way we'd do it is adding a separate jobs table which lists all the job names and their salaries. Then the existing jobs config would just be the default salary if there is no value in the database.
> 
> If someone can get to it before me though that would be great.

might be something we could do with 2.0 and the great db refactor / new group system

Manason:
> > Yeah this is something I'd love to do in the future when I have some time. The way we'd do it is adding a separate jobs table which lists all the job names and their salaries. Then the existing jobs config would just be the default salary if there is no value in the database.
> > If someone can get to it before me though that would be great.
> 
> might be something we could do with 2.0 and the great db refactor / new group system

No need to wait for 2.0. This would be a separate new table. Also has nothing to do with a group system.

Manason:
Attempted to do this, but it was too difficult to maintain backwards compatibility. To do this properly, we'd need to likely remove the jobs.lua entirely and create a set of commands and a UI for modifying the jobs. And we'd need to break QBCore.Shared.Jobs. Will mark this as a v2.0 change

PR #668, opened 2025-01-31
feat(groups): Add group management functionality

citRaTTV:
## Description

Implements external group management via server-sided exports & a DB table, allowing for user-based implementations to exist. Depends on having (new) ConVar `qbx:enableGroupManagement` set to `true` (default `false`).

Closes https://github.com/Qbox-project/qbx_core/issues/235

## Checklist

### DIFF
--- a/client/groups.lua
+++ b/client/groups.lua
@@ -1,16 +1,13 @@
-local jobs = require 'shared.jobs'
-local gangs = require 'shared.gangs'
-
 ---@return table<string, Job>
 function GetJobs()
-    return jobs
+    return lib.callback.await('qbx_core:server:getJobs')
 end
 
 exports('GetJobs', GetJobs)
 
 ---@return table<string, Gang>
 function GetGangs()
-    return gangs
+    return lib.callback.await('qbx_core:server:getGangs')
 end
 
 exports('GetGangs', GetGangs)
@@ -18,6 +15,7 @@ exports('GetGangs', GetGangs)
 ---@param name string
 ---@return Job?
 function GetJob(name)
+    local jobs = lib.callback.await('qbx_core:server:getJobs')
     return jobs[name]
 end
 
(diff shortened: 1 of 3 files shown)


solareon:
So rather than do this in the core this entire functionality should be provided via qbx_management utilizing the existing exports of [CreateJobs](https://docs.qbox.re/resources/qbx_core/exports/server#createjobs) and [CreateGangs](https://docs.qbox.re/resources/qbx_core/exports/server#creategangs) which would prevent any changes in backwards compatibility. The functionality could even be expanded via qbx_adminmenu and qbx_management to enable administrators to define new jobs at runtime.

In either case I recommend a more specific database table name than `groups` as that is likely to be a conflict with other scripts

Additionally the proposed change would greatly increase network traffic instead of caching the results of the Jobs/Gangs table as is done today and having them be updated via a pushed net event.

citRaTTV:
> So rather than do this in the core this entire functionality should be provided via qbx_management utilizing the existing exports of [CreateJobs](https://docs.qbox.re/resources/qbx_core/exports/server#createjobs) and [CreateGangs](https://docs.qbox.re/resources/qbx_core/exports/server#creategangs) which would prevent any changes in backwards compatibility. The functionality could even be expanded via qbx_adminmenu and qbx_management to enable administrators to define new jobs at runtime.
> 
> In either case I recommend a more specific database table name than `groups` as that is likely to be a conflict with other scripts
> 
> Additionally the proposed change would greatly increase network traffic instead of caching the results of the Jobs/Gangs table as is done today and having them be updated via a pushed net event.

Thanks for the review! I decided to implement this in the core as I figured it'd make it easier for others to implement group changes as well, but I'm happy to abandon that in favour of implementing the functionality entirely in qbx_management if that's what's best.

If we do want to keep it in the core, I can work on a caching mechanism rather than relying on callbacks. Let me know which way the maintainers would prefer and I'll adjust 🙂

citRaTTV:
Going to abandon this PR in favour of implementation in qbx_management

PR #676, opened 2025-02-25
Make sure license and check player bucket

nurrifqialhusaini:
## Description

`server/events.lua` just make sure the license its not nil (just make sure)
`server/functions.lua` check if bucket is 0 and using ternary operator

## Checklist


Frowmza:
I see this both useless
- Any case will identifier will be nil?, also why you checking the license and not identifier
- The bucket should never be nil. If someone wants to reset a player's bucket, they should pass 0. Even if we were to accept this, it would not match the function name `SetPlayerBucket`, and we would then be passing a nil bucket.

nurrifqialhusaini:
@Frowmza maybe someone wanna go back to bucket 0 and just destroy source from QBX.Player_Buckets ...
Like maybe make elevator script with bucket system and go back to bucket 0.. but if we `SetPlayerBucket(source, 0)` the QBX.Player_Buckets still saving our data like cached

nurrifqialhusaini:
New Project
Why not destroy when bring back to bucket 0 and just use 1 func not make other func

Frowmza:
Then just check if bucket is 0 then destroy the data, i don't see why we need to allow bucket to be nil. Also, as i said we don't need that identifiers check, it will be never be nil.

nurrifqialhusaini:
thats why i using ternary operator for checking bucket 0

Frowmza:
you're not listening, what I'm trying to clarify, that bucket shouldn't be nil, so if it nil we simply would return. If it passed as 0 then we reset and destroy

nurrifqialhusaini:
@Frowmza So you mean there should be a check to ensure bucket is not nil?

Frowmza:
This will also be a breaking change. However, if `QBX.Player_Buckets` is expected to never be nil, implementing it this way would result in the value being nil. And then break the code of who ever using it. Will hold this until we discuss it with team

nurrifqialhusaini:
@Frowmza Ahhh, I see, the data. Okay, I'll wait for the team's response.. 👍

PR #679, opened 2025-03-04
fix: Add payment protection to pay loop to prevent duplicate paychecks per cycle

binary-knight:
## Description

We had an issue where players were being paid because the pay loop was being called twice externally--after some deep digging and debugging, we weren't able to find the root cause, however, I added some protection against this to the pay loop and it fixed the problem.

## Checklist

### DIFF
--- a/server/loops.lua
+++ b/server/loops.lua
@@ -23,15 +23,23 @@ CreateThread(function()
 end)
 
 local function pay(player)
+    local now = os.time()
+    if player.PlayerData.lastPayCheck and (now - player.PlayerData.lastPayCheck) < (config.money.paycheckTimeout * 60) then
+        return
+    end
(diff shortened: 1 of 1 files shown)


binary-knight:
> I don't see a case where the player can be paid twice in a loop. Are you sure there isn't another conflicting loop?

It's not necessarily the loop paying the player twice, but the function in the loop being called twice, externally.  I have turned off every script, ensured qbx_core was only initialzied once...etc, and it still kept happening.  So---something was calling the loop twice, this fix prevents the player being paid more than once in a time-frame, regardless of how often the loop is called.

jellyton255:
Can't speak for others, but I've never experienced this issue, this might just be something on your end. Why would the pay function be called externally, and how? The pay function is a local unexported function.

PR #681, opened 2025-03-07
feat: live database updates

BerkieBb:
## Description

This PR introduces Live Database Updates. It changes how player data is updated by making sure only the any updated playerdata field is updated specifically in the database each 15 seconds (configurable). This is not directly live, but live enough to ensure data gets updated fast.

I have tested this myself and fixed bugs along but it's best to make sure this gets tested with others as well. However, don't test in production until a few have tested in a test environment.

This also adds 2 new events that will be triggered from SetPlayerData in qbx_core/server/player.lua
qbx_core:client:setPlayerData
qbx_core:server:setPlayerData

## Checklist

### DIFF
--- a/config/server.lua
+++ b/config/server.lua
@@ -1,5 +1,6 @@
 return {
-    updateInterval = 5, -- how often to update player data in minutes
+    updateInterval = 5, -- how often to update player thirst and hunger in minutes
+    dbUpdateInterval = 15, -- how often to update the database for player data in seconds
 
     money = {
         ---@alias MoneyType 'cash' | 'bank' | 'crypto'
--- a/server/functions.lua
+++ b/server/functions.lua
@@ -358,7 +358,7 @@ function ToggleOptin(source)
     if not license or not IsPlayerAceAllowed(source --[[@as string]], 'admin') then return end
     local player = GetPlayer(source)
     player.PlayerData.optin = not player.PlayerData.optin
-    player.Functions.SetPlayerData('optin', player.PlayerData.optin)
+    SetPlayerData(player.PlayerData.source, 'optin', player.PlayerData.optin, nil, true)
 end
 
 exports('ToggleOptin', ToggleOptin)
--- a/server/loops.lua
+++ b/server/loops.lua
@@ -1,4 +1,5 @@
 local config = require 'config.server'
+local storage = require 'server.storage.main'
 
 local function removeHungerAndThirst(src, player)
     local playerState = Player(src).state
@@ -8,8 +9,6 @@ local function removeHungerAndThirst(src, player)
 
     player.Functions.SetMetaData('thirst', math.max(0, newThirst))
     player.Functions.SetMetaData('hunger', math.max(0, newHunger))
-
-    player.Functions.Save()
 end
 
 CreateThread(function()
@@ -19,6 +18,14 @@ CreateThread(function()
         for src, player in pairs(QBX.Players) do
             removeHungerAndThirst(src, player)
         end
+    end
+end)
+
+CreateThread(function()
+    local interval = 1000 * config.dbUpdateInterval
+    while true do
+        Wait(interval)
+        storage.sendPlayerDataUpdates()
     end
 end)
 
--- a/server/storage/players.lua
+++ b/server/storage/players.lua
@@ -1,5 +1,33 @@
 local defaultSpawn = require 'config.shared'.defaultSpawn
 local characterDataTables = require 'config.server'.characterDataTables
+local playerDataUpdateQueue = {}
+local collectedPlayerData = {}
+local isUpdating = false
+local isPlayerUpdating = false
+
+local otherNamedPlayerFields = {
+    ['items'] = 'inventory',
+    ['lastLoggedOut'] = 'last_logged_out'
+}
+
+local jsonPlayerFields = {
+    ['id'] = false,
+    ['userId'] = false,
+    ['citizenid'] = false,
+    ['cid'] = false,
+    ['license'] = false,
+    ['name'] = false,
+    ['money'] = true,
+    ['charinfo'] = true,
+    ['job'] = true,
+    ['gang'] = true,
+    ['position'] = true,
+    ['metadata'] = true,
+    ['inventory'] = true,
+    ['phone_number'] = false,
+    ['last_updated'] = false,
+    ['last_logged_out'] = false
+}
 
 local function createUsersTable()
     MySQL.query([[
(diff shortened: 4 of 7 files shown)


server/storage/players.lua
```diff
@@ -1,5 +1,33 @@
...
+    ['charinfo'] = true,
+    ['job'] = true,
+    ['gang'] = true,
+    ['position'] = true,
+    ['metadata'] = true,
+    ['inventory'] = true,
+    ['phone_number'] = false,
+    ['last_updated'] = false,
+    ['last_logged_out'] = false
+}
```
CodexisPhantom, line 30: I don't know if I'm right, but I think you could specify only the json fields.

BerkieBb: Yeah I'm not sure why I did it like that

BerkieBb: Oh nevermind, I just remembered. I did it like that to also acts as a recognition for database columns. If the key is not in that list, it's not valid or it's supposed to be in a table.

BerkieBb: You can see that on line 425 of that file

CodexisPhantom: Oh okey I see then I said nothing

ihyajb:
Tested these changes solo and ran into no issues. Will test with some more people later tomorrow

tomiichx:
I'm not really a contributor, but I'm using `qbx_core`, and I enjoy checking out the latest changes/PRs from time to time 😄

I really like the direction of this PR, but I wanted to suggest one potential improvement. I noticed you're currently looping through the gathered player data and making individual DB calls for each iteration.

As a suggestion, implementing batch saving and wrapping it in a transaction would both improve performance and add safety - ensuring that if any errors occur, nothing gets partially saved (with errors being logged, of course). Batch saving would be significantly more efficient rather than executing queries one by one.

You might also consider going one step further and define a max batch size, and if it exceeds X (e.g. 100), you run an another save function for the rest of the data. This might be overkill though and will most probably never happen 😄 

Just a thought that might help make this PR even better

BerkieBb:
> I'm not really a contributor, but I'm using `qbx_core`, and I enjoy checking out the latest changes/PRs from time to time 😄
> 
> I really like the direction of this PR, but I wanted to suggest one potential improvement. I noticed you're currently looping through the gathered player data and making individual DB calls for each iteration.
> 
> As a suggestion, implementing batch saving and wrapping it in a transaction would both improve performance and add safety - ensuring that if any errors occur, nothing gets partially saved (with errors being logged, of course). Batch saving would be significantly more efficient rather than executing queries one by one.
> 
> You might also consider going one step further and define a max batch size, and if it exceeds X (e.g. 100), you run an another save function for the rest of the data. This might be overkill though and will most probably never happen 😄
> 
> Just a thought that might help make this PR even better

My thought was actually that it might be better to do individual PRs so atleast something gets saved. So when one little thing fails, a server owner can catch in the console and either manually update it or check what's wrong/report it. But at the same time now that I think about it, it's not wrapped to catch errors so it'd stop execution which is not good.

tomiichx:
Hello again, @BerkieBb!

I had to come back here with a PR related question;
We have a jobscreator on our server which allows us to quickly create jobs in-game (Lunar)

However, some players after relogging simply aren't employed somehow, or atleast they don't see the stash/boss menu interaction markers.

I figured the fix could be to switch to ox_target because it always works (don't ask me why or how, it just works), but I'm still not comfortable with having this issue on the LIVE server, because I'm unsure if it's related to lunar_bridge interactions, or a general core bug, and could this PR be the fix?

Also, I'm on the most recent version of the core with no modifications related to groups

Thanks in advance!

BerkieBb:
> Hello again, @BerkieBb!
> 
> I had to come back here with a PR related question; We have a jobscreator on our server which allows us to quickly create jobs in-game (Lunar)
> 
> However, some players after relogging simply aren't employed somehow, or atleast they don't see the stash/boss menu interaction markers.
> 
> I figured the fix could be to switch to ox_target because it always works (don't ask me why or how, it just works), but I'm still not comfortable with having this issue on the LIVE server, because I'm unsure if it's related to lunar_bridge interactions, or a general core bug, and could this PR be the fix?
> 
> Also, I'm on the most recent version of the core with no modifications related to groups
> 
> Thanks in advance!

This PR won't be the fix for you. Current core does the job well, so it might be an issue with another resource. Also, this should've been posted as a support question in the discord, not on this PR.

config/server.lua
```diff
@@ -1,5 +1,6 @@
 return {
-    updateInterval = 5, -- how often to update player data in minutes
+    updateInterval = 5, -- how often to update player thirst and hunger in minutes
+    dbUpdateInterval = 15, -- how often to update the database for player data in seconds
```
YvanCywan, line 3: The variable names should be changed to become more descriptive instead of having the comments here.

Stuff like, ```attributeUpdateIntervalMinutes``` and ```dbUpdateIntervalMinutes``` are both descriptive and completely reduce the need to comment.

Commenting should always encapsulate why instead of how.

server/functions.lua
```diff
@@ -358,7 +358,7 @@ function ToggleOptin(source)
     if not license or not IsPlayerAceAllowed(source --[[@as string]], 'admin') then return end
     local player = GetPlayer(source)
     player.PlayerData.optin = not player.PlayerData.optin
-    player.Functions.SetPlayerData('optin', player.PlayerData.optin)
+    SetPlayerData(player.PlayerData.source, 'optin', player.PlayerData.optin, nil, true)
```
YvanCywan, line 361: Never really a fan of passing in nil/null values into methods, rife for ungraceful exiting.

server/storage/players.lua
```diff
@@ -382,6 +410,124 @@ local function cleanPlayerGroups()
...
+                if not current[key[i]] then
+                    current[key[i]] = {}
+                end
+
+                current = current[key[i]]
+            end
+        end
+
+        current[key[#key]] = value
+    end
```
YvanCywan, line 458: This could be reworked to reduce nesting

server/storage/players.lua
```diff
@@ -382,6 +410,124 @@ local function cleanPlayerGroups()
...
+    if not currentTable[citizenid] then
+        currentTable[citizenid] = {}
+    end
+
+    currentTable[citizenid][hasSubKeys and key[1] or key] = hasSubKeys and {} or value
+
+    if hasSubKeys then
+        local current = currentTable[citizenid][key[1]]
+        if #key > 2 then
+            -- We don't check the last one because otherwise we lose the table reference
```
YvanCywan, line 447: Example of really good commenting.

server/storage/players.lua
```diff
@@ -382,6 +410,124 @@ local function cleanPlayerGroups()
...
+
+    collectedPlayerData = playerDataUpdateQueue
+    playerDataUpdateQueue = {}
+    isUpdating = false
+end
+
+---@param citizenid string
+local function forcePlayerDataUpdate(citizenid)
+    -- We don't need to update a single player when everyone is already getting an update
+    if isUpdating then return end
```
YvanCywan, line 507: If this isn't self explanatory then some refactoring is necessary.

server/loops.lua
```diff
@@ -22,6 +21,14 @@ CreateThread(function()
     end
 end)
 
+CreateThread(function()
+    local interval = 1000 * config.dbUpdateInterval
```
YvanCywan, line 25: So it's in minutes, but you're just multiplying the value by 1000?

Considering the flow of the update is single threaded, if anything holds up the update, the timer won't start until the previous operation has finished executing.

To make the timer more accurate, you should be using a clock implementation.

YvanCywan (commented):
Hey, been a while! I know. Saw this PR on my feed and wanted to check it out.

Some pretty decent stuff here.

Also fairly unsure if "live" is the correct way to phrase these db updates. Although, I am less picky about that since anything compared to how it was is relatively "live".

config/server.lua
BerkieBb: Considering there are no other variables that could represent something like this, a shorter name is good to make the code more readable. I don't like long names for variables.

server/functions.lua
BerkieBb: Couldn't find another way to keep backwards compatibility

server/storage/players.lua
BerkieBb: Not if you want to keep it readable

server/storage/players.lua
BerkieBb: You underestimate how much people can understand code and who can view this code.

server/loops.lua
BerkieBb: > So it's in minutes, but you're just multiplying the value by 1000?

Good catch, didn't notice that.

> Considering the flow of the update is single threaded, if anything holds up the update, the timer won't start until the previous operation has finished executing.
>
> To make the timer more accurate, you should be using a clock implementation.

The holdup time here is incredibly minimal and lua is incredibly fast here. So no, this is good.

BerkieBb: > Good catch, didn't notice that.

Nevermind this, I didn't read my own comment. The db update interval is in seconds, not minutes. To keep it relatively live.

BerkieBb: Yes it might be a bit confusing, but both of them handle different things. The update interval updates hunger and thirst only and handles how much hunger and thirst you lose in what time, people used minutes always there.

The db update handles most recent update tasks and we can do it in seconds because it's fast but not too fast to cause any issues when having huge queries.

server/storage/players.lua
YvanCywan: Your dealing with the edge case in the first if statement, inverting the first statement into a guard clause will definitely improve readability.

server/loops.lua
YvanCywan: > The holdup time here is incredibly minimal and lua is incredibly fast here. So no, this is good.
This argument only works if everyone is using the same infra and nothing else impacts it. A clock ensures better consistency and resiliency. You should always design stuff to be as resilient as possible, not "this is good enough".

Seconds is fine as well, no argument against leaving this in seconds.

> Nevermind this, I didn't read my own comment. The db update interval is in seconds, not minutes. To keep it relatively live
Hence why it should be in the variable name. More descriptive naming is usually better, adding a word or two here could do wonders to help differentiate them.

> Yes it might be a bit confusing, but both of them handle different things. The update interval updates hunger and thirst only and handles how much hunger and thirst you lose in what time, people used minutes always there.

Again why it should be reflected better in variable naming, see how easy it was for me to mix up which one was meant to be minutes and the other one is meant to be seconds?

server/functions.lua
YvanCywan: I understand, sucks dick tho

server/storage/players.lua
BerkieBb: Done

PR #689, opened 2025-04-21
feat(commands): add opt-in command & enforce opt-in for admin commands

its4lion:
## Description
This PR adds opt-in checks for admin commands `tp`, `tpm` etc,
Admins must opt-in to use them,New `optin` command toggles it. Prevents accidental admin actions.

## Checklist

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -21,14 +21,16 @@
         "connecting_error": "An error occurred while connecting to the server. (Check your server console)",
         "no_match_character_registration": "Anything other than letters aren't allowed, trailing whitespaces aren't allowed either and words must start with a capital letter in input fields. You can however add words with spaces inbetween.",
         "already_in_queue": "You are already in queue.",
-        "no_subqueue": "You were not let in any sub-queue."
+        "no_subqueue": "You were not let in any sub-queue.",
+        "not_optin": "You are not opted in."
     },
     "success": {
         "server_opened": "The server has been opened",
         "server_closed": "The server has been closed",
         "teleported_waypoint": "Teleported To Waypoint.",
         "character_deleted": "Character deleted!",
-        "character_deleted_citizenid": "You successfully deleted the character with Citizen ID %s."
+        "character_deleted_citizenid": "You successfully deleted the character with Citizen ID %s.",
+        "optin_set": "Opt in has been set to %s successfully."
     },
     "info": {
         "received_paycheck": "You received your paycheck of $%s",
@@ -170,6 +172,7 @@
             "params": {
                 "message": { "name": "message", "help": "Message to send" }
             }
-        }
+        },
+        "optin": { "help": "Toggle opt-in (Admin Only)" }
     }
 }
--- a/server/commands.lua
+++ b/server/commands.lua
@@ -12,6 +12,7 @@ lib.addCommand('tp', {
     },
     restricted = 'group.admin'
 }, function(source, args)
+    if not IsOptin(source) then return Notify(source, locale('error.not_optin'), 'error') end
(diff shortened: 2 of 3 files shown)


server/commands.lua
```diff
@@ -12,6 +12,7 @@ lib.addCommand('tp', {
     },
     restricted = 'group.admin'
 }, function(source, args)
+    if not IsOptin(source) then return Notify(source, locale('error.not_optin'), 'error') end
```
BerkieBb, line 15: Put an empty line under the returns for readability. Do this for all of the commands.

server/commands.lua
```diff
@@ -40,13 +41,15 @@ lib.addCommand('tpm', {
...
     TriggerClientEvent('QBCore:Command:GoToMarker', source)
 end)
 
 lib.addCommand('togglepvp', {
     help = locale('command.togglepvp.help'),
     restricted = 'group.admin'
-}, function()
+}, function(source)
+    if not IsOptin(source) then return Notify(source, locale('error.not_optin'), 'error') end
```
BerkieBb, line 52: Put an empty line under this one for readability.

server/commands.lua
t0r3tto: Still unresolved, https://github.com/Qbox-project/qbx_core/pull/689/commits/16e8c9278703fc7844b13e0776e023c51c582541 doesn't address Berkie's feedback; this is how it should look like to improve readability:



```suggestion
}, function(source, args)
    if not IsOptin(source) then Notify(source, locale('error.not_optin'), 'error') return end

    if args[locale('command.tp.params.x.name')] and not args[locale('command.tp.params.y.name')] and not args[locale('command.tp.params.z.name')] then
```

locales/en.json
```diff
@@ -22,14 +22,16 @@
...
+        "not_optin": "You are not opted in."
     },
     "success": {
         "server_opened": "The server has been opened",
         "server_closed": "The server has been closed",
         "teleported_waypoint": "Teleported To Waypoint.",
         "character_deleted": "Character deleted!",
-        "character_deleted_citizenid": "You successfully deleted the character with Citizen ID %s."
+        "character_deleted_citizenid": "You successfully deleted the character with Citizen ID %s.",
+        "optin_set": "Opt in has been set to %s successfully."
```
t0r3tto, line 34: "You have successfully opted in/out" is clearer than on/off imo

t0r3tto (changes requested):
**Should be good after the suggested changes.**

Also, it might be a good idea to make the Opt In/Out functionality configurable - disabled by default - since it's never really been used before. That way, people can enable it manually if they want, rather than having it introduced abruptly.

its4lion:
> **Should be good after the suggested changes.**
> 
> Also, it might be a good idea to make the Opt In/Out functionality configurable - disabled by default - since it's never really been used before. That way, people can enable it manually if they want, rather than having it introduced abruptly.

i dont know really how we should do that 

maybe always return true at `IsOptin` function and disable the command optin ?

t0r3tto:
> > **Should be good after the suggested changes.**
> > Also, it might be a good idea to make the Opt In/Out functionality configurable - disabled by default - since it's never really been used before. That way, people can enable it manually if they want, rather than having it introduced abruptly.
> 
> i dont know really how we should do that
> 
> maybe always return true at `IsOptin` function and disable the command optin ?

Yep, if the whole "Opt In/Out" functionality is disabled then always return true in `IsOptin`.
That was a suggestion, maybe for another PR; but AFAIC this PR is good to go.

t0r3tto (approved):
LGTM

PR #691, opened 2025-05-09
fix nil error on delete char

ggfto:
## Description

This is just a small fix for an error on delete a character, the error is shown like this, server side:
image

## Checklist

- [X] My pull request fits the contribution guidelines & code conventions. (At least I hope)


Frowmza:
I don't get the purpose of this PR.
First the line in the error you provided it not the same you edited.
Second is you did nothing, the change will have same result, `.license` notation is same as `['license']`, the difference is we use dot notation because we know the table property name, and bracket notation used for dynamic.

PR #692, opened 2025-05-09
fix(server): use callbacks for character deletion

t0r3tto:
## Description

This PR fixes the famous delete character bug, the implemented fix replaces the internal function `DeleteCharacter` by a callback `qbx_core:server:deleteCharacter` ; ensuring data is well received before fetching displaying nonexistent characters.

The notification is handled client-side based on the return value, instead of the server

_Error preview:_
image

## Checklist

- [X] I have personally loaded this code into an updated Qbox project and checked all of its functionality.
- [X] My pull request fits the contribution guidelines & code conventions.

PR #693, opened 2025-05-09
fix(server/player): fetch player check

Frowmza:
## Checklist

### DIFF
--- a/server/player.lua
+++ b/server/player.lua
@@ -1424,7 +1424,8 @@ exports('GetMoney', GetMoney)
 ---@param citizenid string
 function DeleteCharacter(source, citizenid)
     local license, license2 = GetPlayerIdentifierByType(source --[[@as string]], 'license'), GetPlayerIdentifierByType(source --[[@as string]], 'license2')
-    local result = storage.fetchPlayerEntity(citizenid)?.license
+    local playerEntity = storage.fetchPlayerEntity(citizenid)
+    local result = playerEntity and playerEntity.license
     if license == result or license2 == result then
         CreateThread(function()
             local success = storage.deletePlayer(citizenid)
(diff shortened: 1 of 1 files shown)


server/player.lua
```diff
@@ -1424,7 +1424,7 @@ exports('GetMoney', GetMoney)
 ---@param citizenid string
 function DeleteCharacter(source, citizenid)
     local license, license2 = GetPlayerIdentifierByType(source --[[@as string]], 'license'), GetPlayerIdentifierByType(source --[[@as string]], 'license2')
-    local result = storage.fetchPlayerEntity(citizenid).license
+    local result = storage.fetchPlayerEntity(citizenid)?.license
```
Ehbw, line 1427: Wouldn't use safe navigation anymore. 99% of the time its fine but there has been an increase in reports of it failing and providing an broken stacktrace

Frowmza: Agree, it failing for me to sometimes but only when i use it after a bracket notation `[random]?.random` or `[random]?[random2]`

t0r3tto:
Hello @Frowmza ,

Closing my PR https://github.com/Qbox-project/qbx_core/pull/692 without any explanation or valid reasoning is unacceptable. I’ve tested your 2 lines  "fix" (or shall I say characters?), it does absolutely nothing. A video recording demonstrating your "fix" is linked at the end of this comment, feel free to check it out.

If you are going to dismiss someone else's work, you should at least provide an actual solution, instead of useless unrelated changes. Your approach here undermines the project and the collective effort, especially for a community driven project. 

While there are no rules requiring you to wait for more reviews, actions like this should involve additional reviewers before being made to ensure the decision is correct.

Video demonstration: https://www.youtube.com/watch?v=ZbulDgCH3hA


Thank you.

Frowmza:
I was writing the reason but got distracted. My fix checks for the actual error, which is the main goal of both your PR and mine. From the video, it looks like the characters didn’t refresh, on the second click, you got kicked because the character didn’t exist anymore, and the system saw it as an exploit.
Your PR doesn't make any sense, we don't need a callback and you calling a server callback in server, also in client where you used a ternary to return a string without wrapping it in a function. That looks rushed and shows a lack of experience.

Frowmza:
ok my bad, i've reopened you PR

PR #700, opened 2025-05-27
fix: qb-spawn event

zeddosx:
## Description

`qb-spawn:client:openUI` doesnt really exist in qbx-spawn, and that causes an issue when you try to run a chat like okok for example. Calling the function instead makes it work, I am pretty sure that it is thanks to calling the exports.spawnmanager:spawnPlayer function. The event called before the function updates the coords you choose on the qbx-spawn map so it always spawns you in the right place. 

## Checklist

- [ x] I have personally loaded this code into an updated Qbox project and checked **the spawn functionality**.

### DIFF
--- a/client/character.lua
+++ b/client/character.lua
@@ -287,7 +287,7 @@ local function spawnDefault() -- We use a callback to make the server wait on th
     TriggerEvent('qb-clothes:client:CreateFirstCharacter')
 end
 
-local function spawnLastLocation()
+local function spawnLocation()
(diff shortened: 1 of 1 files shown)


D4isDAVID:
What kind of issue are you having? These changes will constantly spawn the player in their last location. `qb-spawn:client:openUI` is provided purely for compatibility and instead `qb-spawn:client:setupSpawns` is the event with the actual functionality.

PR #716, opened 2025-09-23
fix(server/player): Missing bankAuth in SetPlayerPrimaryJob -> toPlayerJob

stone9k:
## Description

bankAuth was not set when setting job with SetPlayerPrimaryJob -> toPlayerJob

## Checklist


Frowmza:
imo, static data shouldn't be in player data at all, bankAuth, isboss, payment, label. We could always access them thru core shared config/export by key which is player job name, so I think adding this is useless.

PR #724, opened 2025-11-30
feat(discord): show player name in rich presence

Klaus-Michealson:
Adds the connected player's Steam name to Discord Rich Presence.

Before:
→ Players 42/128

After:
→ HexaN | 42/128

This gives a more personal and modern look that many servers already use. The change is fully optional (works only if Discord Rich Presence is enabled) and does not affect performance.

Tested on latest Qbox version, no errors.

## Checklist


Kenshiin13 (changes requested):
Can you use config-driven placeholders instead and keep the default behaviour? So every server can decide for themselves what they wish to show in Rich Presence.

`config.richPresence = "{steamName} | {currentPlayers}/{maxPlayers}"`

KlausMichealson-Dev:
Yes, I can, that would be a good idea.

Kenshiin13:
> Yes, I can, that would be a good idea.

Already commited

Klaus-Michealson:
> > Evet, yapabilirim, bu iyi bir fikir olurdu.
> 
> Zaten taahhütte bulundum.

Hello, yes, I haven't been active on my account for a few days. I will fulfill your requests, please give me some time.

PR #726, opened 2025-12-05
Add config setting for /optin admin menu command

nzkfc:
## Description

Adds a configuration toggle to skip the /optin command. Searching "Optin" on the Qbox Discord shows many support answers explaining its purpose, this setting will remove most of the "Why can't I use commands" questions.

While the original implementation was useful, server owners need more control. This change avoids the need to modify qbx_core directly, as well as having to maintain it with each update and gives owners better control over its behavior.

I'm not sure if there was a specific reason for the original change to add this in too deem it mandatory, so let me know if there was one.

## Checklist

- [X] I have no errors in console and I have run through all commands to ensure they work.
- [X] I love ducks, quack!

### DIFF
--- a/server/functions.lua
+++ b/server/functions.lua
@@ -343,6 +343,7 @@ exports('GetPermission', GetPermission)
 ---@param source Source
 ---@return boolean
 function IsOptin(source)
+    if not serverConfig.requireOptIn then return true end
     local license = GetPlayerIdentifierByType(source --[[@as string]], 'license2') or GetPlayerIdentifierByType(source --[[@as string]], 'license')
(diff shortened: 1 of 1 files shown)


server/commands.lua
```diff
@@ -12,7 +12,7 @@ lib.addCommand('tp', {
     },
     restricted = 'group.admin'
 }, function(source, args)
-    if not IsOptin(source) then Notify(source, locale('error.not_optin'), 'error') return end
+    if config.server.requireOptIn and not IsOptin(source) then Notify(source, locale('error.not_optin'), 'error') return end
```
its4lion: better to edit the main function `IsOptin` so other scripts can work fine with it

PR #734, opened 2026-01-11
feat: Add cooldown in lib.lua

SUWAROW:
## Description

Stop flood request to server

- Added cooling helpers to `qbx_core/modules/lib.lua`:
  - `qbx.isCooldownReady` tracks per-key cooldowns and optionally notifies players, with built-in flood control.
  - `qbx.setCooldown` allows manual timestamp updates when needed.
  - Inline docs describe parameters and defaults.

## Checklist

### DIFF
--- a/modules/lib.lua
+++ b/modules/lib.lua
@@ -704,31 +704,4 @@ else
     end
 end
 
-local _cooldowns = {}
-local function getTimestampMs()
-    if GetGameTimer then
-        return GetGameTimer()
-    end
-    return os.time() * 1000
-end
-
----@param key string
----@param durationMs number?
----@return boolean ready
-function qbx.isCooldownReady(key, durationMs)
-    local now = getTimestampMs()
-    local expires = _cooldowns[key]
-    if not expires or now >= expires then
-        _cooldowns[key] = now + (durationMs or 0)
-        return true
-    end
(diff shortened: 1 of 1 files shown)


Kenshiin13:
- No player scoping. `_cooldowns` is keyed by a raw global string, so on the
  server `isCooldownReady('xyz', 5000)` puts every player on cooldown at once.
- `isCooldownReady` mutates. A `true` return immediately starts the timer. The name implies a pure read.
- `qbx.setCooldown` is missing. The description lists it but it isn't in the diff,
  so the API is incomplete.
- `_cooldowns` / `_notifyTimestamps` are never cleaned up
- The notification makes no sense. Client-side `lib.notify` works fine. Server-side
  `lib.notify` is also truthy because ox_lib's loader assigns a noop when a module fails to load, so the `if lib and lib.notify` branch is always taken and silently does nothing. That makes the `elseif not isServer` branch dead, which doesnt make sense to begin with. And even if it did, it calls the `qbx_core:Notify` export, which does not exist server-side and is given no net id. So it never notifies on the server in either path.

It's clear this was never tested, which on its own is reason to close. In any case, a cooldown helper is general-purpose utility that belongs in ox_lib, not in qbx_core.

PR #737, opened 2026-03-11
stop the error on player join denial. (banned from other script, whitelist/blacklist script, txadmin whitelist pre-check, etc...)

Somis12:


Kenshiin13 (changes requested):
Please add a description outlining what this fixes and its implications. It's currently unclear what problem this addresses.

PR #744, opened 2026-05-03
feat: fire QBCore:Server:PayCheck server event after paycheck

antonstjernquist:
## Problem

Third-party resources (phone mods, banking overlays, job trackers) currently have no way to react to a player receiving their paycheck without modifying `config/server.lua` themselves. The `sendPaycheck` config function is a customisation point for *how* to pay, not a hook for *observing* that a payment happened.

## Solution

Fire a server-local event after every successful paycheck:

```lua
TriggerEvent('QBCore:Server:PayCheck', source --[[ number ]], payment --[[ number ]])
```

This uses `TriggerEvent` (not `TriggerNetEvent`) so it stays server-side and is zero-cost when nothing is listening.

The event name `QBCore:Server:PayCheck` is already used as the de-facto convention by several community resources that fire it themselves — this makes qbx_core the canonical source.

## Changes

- `server/loops.lua` — refactored `pay()` to a single `sendPaycheck` call site (eliminating three duplicated early-return paths) and fires the event afterwards.

No config changes, no breaking changes.


Kenshiin13 (changes requested):
1. Move the emit into config.sendPaycheck ([config/server.lua:138](https://github.com/Qbox-project/qbx_core/blob/97c91b091de07bc24436d96798e786ceadb12b7e/config/server.lua#L138)) instead of the three pay() call sites. It's the single point a paycheck is issued, so it fires once and stays correct if a pay path is ever added.

2. QBCore:Server:* is the legacy bridge namespace. Use qbx_core:server:* with the repo's on<Event> convention (onJobUpdate, onGangUpdate, onSetMetaData) → qbx_core:server:onPaycheck.

Kenshiin13 (approved):

PR #756, opened 2026-06-07
index user lookups and stop the per-interval loop hitch

WhereiamL:
## Description

`fetchUserByIdentifier` resolves a connecting player with
`SELECT userId FROM users WHERE license = ?` (or license2 / fivem / discord),
but the `users` table only has a primary key on `userId` — none of the
identifier columns are indexed. Since `users` grows for every player that ever
joined, that lookup becomes a full table scan and the connect deferral stalls
for ~1s on a server with a large table.

This adds the missing indexes, and while here cleans up two getters/loops that
scale poorly with player count.

### Changes

- Index `users` on `license`, `license2`, `fivem`, `discord` in
  `createUsersTable`, using `CREATE INDEX IF NOT EXISTS` so existing databases
  pick them up on resource start. Connect lookup goes from a full scan to an
  index seek.
- Add a reverse lookup registry (citizenid / userId / identifier → source) kept
  up to date on player load/unload, so `GetSource`, `GetUserId`,
  `GetPlayerByCitizenId` and `GetPlayerByUserId` resolve in constant time
  instead of scanning every online player. Fast paths revalidate and fall back
  to a scan, so they can't return stale data.
- Process the hunger/thirst and paycheck loops in batches with a frame yield
  instead of iterating every player on a single frame. The player list is
  snapshotted first so a join/drop during a yield can't corrupt iteration.

No API or behaviour changes.

## Checklist

### DIFF
--- a/server/character.lua
+++ b/server/character.lua
@@ -51,6 +51,43 @@ lib.callback.register('qbx_core:server:loadCharacter', function(source, citizenI
     lib.print.info(('%s (Citizen ID: %s ID: %s) has successfully loaded!'):format(GetPlayerName(source), citizenId, source))
 end)
 
+--- Validates client-supplied character info. Never trust the values the create
+--- screen sends: cap free-text length, enforce types, and reject anything that
+--- isn't a usable string so a crafted payload can't bloat the row or smuggle in
+--- non-string fields. Returns nil to refuse creation outright.
+---@param data table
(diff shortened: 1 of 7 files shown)


ChatDisabled:
the index changes on users are "sure" I guess. We're talking about milliseconds here and it requires a bump in our mariadb requirement. But 🤷

PR #761, opened 2026-06-12
fix(server/loops): guard paycheck against missing job grade

WhereiamL:
pay() indexed GetJob(job.name).grades[job.grade.level].payment with no nil checks. A player whose job or grade no longer exists in shared jobs (removed/renamed job, edited grades) threw an error inside the paycheck loop, which runs unprotected in a while-true thread. That killed the thread and stopped paychecks for every player until a restart.

Resolve the job and grade defensively, log and skip the player when the grade is gone, and reuse the resolved job table for the offDutyPay check instead of looking it up a second time.

## Checklist

### DIFF
--- a/server/loops.lua
+++ b/server/loops.lua
@@ -24,9 +24,15 @@ end)
 
 local function pay(player)
     local job = player.PlayerData.job
-    local payment = GetJob(job.name).grades[job.grade.level].payment or job.payment
+    local jobData = GetJob(job.name)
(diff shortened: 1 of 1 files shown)


ChatDisabled:
don't know how this would happen since afaik upon connect missing grades/jobs get reset for the player. but no doubt people could probably do something stupid with the runtime job editing functions

PR #766, opened 2026-07-14
fix: assign correct cid when creating a new character

RSR10-fivem:
## Summary
- `createCharacter` never set `charinfo.cid` on a newly created character, because `sanitizeNewCharInfo()` strips the client-supplied `cid` field (correctly, since client input shouldn't be trusted for this).
- `CheckPlayerData` then falls back to `playerData.charinfo?.cid or playerData.cid or 1`, so every new character on an account silently gets `cid = 1`, regardless of how many characters already exist.
- This means a player's 2nd, 3rd, etc. character all end up with the same `cid`, instead of an incrementing per-account slot number.

## Fix
Compute the next `cid` server-side from the account's existing characters (`max(existing cids) + 1`), reusing the character list that's already fetched for the character-limit check in `createCharacter`. Using `max + 1` rather than `count + 1` avoids collisions if a character in the middle of the sequence was previously deleted.

## Test plan
- [ ] Create a fresh account, create 2+ characters, confirm `cid` in the `players` table increments (1, 2, 3, ...) instead of staying at 1
- [ ] Delete a character mid-sequence (e.g. cid 2 of 1/2/3) and create a new one, confirm it gets cid 4 (no collision with existing cid 3)
- [ ] Confirm character list ordering (`ORDER BY cid`) in the multichar UI still reflects creation order

### DIFF
--- a/server/character.lua
+++ b/server/character.lua
@@ -27,6 +27,10 @@ end
 
 lib.callback.register('qbx_core:server:getCharacters', function(source)
     local license2, license = GetPlayerIdentifierByType(source, 'license2'), GetPlayerIdentifierByType(source, 'license')
+    -- Heal any account left with a broken cid sequence (duplicates from the
+    -- pre-fix createCharacter bug, or gaps from deleted characters) before
+    -- handing the list to the client, so create-time logic can stay a simple append.
+    storage.normalizeCids(license2, license)
     return storage.fetchAllPlayerEntities(license2, license), getAllowedAmountOfCharacters(license2, license)
 end)
 
--- a/server/character.lua
+++ b/server/character.lua
@@ -88,14 +88,11 @@ local function sanitizeNewCharInfo(data)
     }
 end
 
----@param existingCharacters PlayerEntity[]
+---@param existingCharacters PlayerEntity[] sorted ascending by cid (storage.fetchAllPlayerEntities orders by cid)
 ---@return integer
 local function getNextCid(existingCharacters)
-    local nextCid = 0
-    for i = 1, #existingCharacters do
-        nextCid = math.max(nextCid, existingCharacters[i].charinfo.cid or 0)
-    end
-    return nextCid + 1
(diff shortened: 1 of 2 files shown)


Kenshiin13 (changes requested):
1) Please revert e006bcf and keep b20afcc's `last + 1`. The backfill collides on the
exact data this PR targets: an account with several pre-fix characters all at cid 1
gives char 4 -> 2 (ok), then char 5 -> 2 again (list is {1,1,1,2}, first mismatch is
still index 2), colliding, and every later char also gets 2. Append-after-highest is
collision-safe regardless of duplicates.

2) Bigger picture, and I think the better direction: the real problem is corrupted
legacy data (many characters at cid 1), and neither append nor backfill heals it, they
only avoid adding to it. Instead of deriving the next cid at create time, normalize an
account's cids when its character list is loaded (the getCharacters step), independent
of creating a new character:

- When fetching the account's characters, if their cids aren't a clean 1..N (duplicates
  or gaps), reassign 1..N and persist it (cid column + the charinfo mirror).
- If they're already 1..N, do nothing, so it only writes once.
- Reassign in a stable order (creation / row order) so a character keeps its slot across
  logins instead of shuffling.

Once cids are guaranteed clean, the create-time logic becomes trivial and safe (append,
or backfill if you want to reuse deleted slots), and existing accounts actually get
healed instead of carrying the duplicate 1s forever.

Since (2) is a bigger change than this PR, it may be worth splitting: land the minimal
`last + 1` fix here, and do the load time normalization as a follow-up.

RSR10-fivem:
Added the load-time normalization we discussed as a second commit here instead of a separate PR — `storage.normalizeCids()` reassigns an account's cids to a clean 1..N sequence (creation order) in `getCharacters` whenever they aren't already clean, so create-time `last + 1` stays safe against legacy duplicate-1 data. Ready for another look.

Kenshiin13 (approved):
LGTM

PR #767, opened 2026-07-14
fix: normalize account cids to a clean 1..N sequence on load

RSR10-fivem:
## Summary
Follow-up to #766, per @Kenshiin13's review there:

> the real problem is corrupted legacy data (many characters at cid 1), and neither append nor backfill heals it, they only avoid adding to it. Instead of deriving the next cid at create time, normalize an account's cids when its character list is loaded (the getCharacters step), independent of creating a new character

- `storage.normalizeCids(license2, license)` fetches an account's characters ordered by `id` (creation order) and checks whether `cid` already forms a clean `1..N` sequence.
- If it does, nothing is written.
- If it doesn't (duplicates from the pre-#766 bug, or gaps left by deleted characters), it reassigns `cid = 1..N` in creation order and persists only the rows that changed (`cid` column + the `charinfo` JSON mirror).
- Called from `qbx_core:server:getCharacters` before the character list is returned to the client, so an account is healed the next time it opens the character-select screen — independent of ever creating a new character.

With this in place, #766's create-time `last + 1` logic is safe to reason about: by the time `createCharacter` runs, cids are already guaranteed clean, so a plain append can't collide or perpetuate old damage.

## Test plan
- [x] Seed an account with several characters all at `cid = 1` (the pre-#766 bug state), call `getCharacters`, confirm cids get rewritten to `1, 2, 3, ...` in original creation (`id`) order and the change persists across a second call (no-op, no extra writes)
- [ ] Seed an account with a gap (e.g. `1, 2, 5` from a deleted character), confirm it normalizes to `1, 2, 3`
- [ ] Seed an already-clean account (`1, 2, 3`), confirm `getCharacters` performs no UPDATE queries
- [ ] Create a new character after normalization, confirm it gets `cid = N + 1` via #766's logic with no collision

### DIFF
--- a/server/character.lua
+++ b/server/character.lua
@@ -27,6 +27,10 @@ end
 
 lib.callback.register('qbx_core:server:getCharacters', function(source)
     local license2, license = GetPlayerIdentifierByType(source, 'license2'), GetPlayerIdentifierByType(source, 'license')
+    -- Heal any account left with a broken cid sequence (duplicates from the
+    -- pre-fix createCharacter bug, or gaps from deleted characters) before
+    -- handing the list to the client, so create-time logic can stay a simple append.
+    storage.normalizeCids(license2, license)
     return storage.fetchAllPlayerEntities(license2, license), getAllowedAmountOfCharacters(license2, license)
 end)
 
--- a/server/character.lua
+++ b/server/character.lua
@@ -88,11 +88,17 @@ local function sanitizeNewCharInfo(data)
     }
 end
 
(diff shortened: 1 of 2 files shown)


RSR10-fivem:
@Kenshiin13 opened this as the follow-up we discussed on #766 — normalizing an account's cids to a clean 1..N sequence on load instead of only handling it at create time. Would appreciate a look when you have a chance.

RSR10-fivem:
Closing — folding this into the #766 thread instead of splitting into a separate PR.

PR #768, opened 2026-07-16
fix(groups): normalize numeric grade keys

### LINKED ISSUE #543: CreateJob/CreateGang don't validate grades are numbers
### Summary

CreateJob/CreateGang don't validate grades are numbers

### Reproduction

Use bridge layer with a script designed to add jobs/gangs at run time with qb-core style grades and they are inserted into the job/gang tables with string grade keys vice numeric ones.

### Expected behavior

That grades are checked to be numeric before inserting into the table

### Actual behavior

The grades are inserted as is without validation

### Additional context

The code in question is in `server/groups.lua` which just requires some additional sanity checks before inserting into the group grades table.

Unsure if the grade data returned via the bridge layer functions should translate the grades back into strings or if this can be safely ignored.

### Current Version

v1.18.0

### Custom Resources

n/a

pushkart2:
## Summary

- convert QB-style numeric string grade keys (for example, `"0"`) to integer keys before caching jobs and gangs
- reject negative, fractional, non-numeric, and colliding grade keys
- apply the same normalization to job and gang grade upserts
- validate gang collections before mutating the shared cache
- acknowledge the two valid `qbx.math.isFinite` calls until the external `qbox_lib` Luacheck standard includes that helper

## Why

Qbox grade lookups use numeric levels, while QB-compatible runtime group definitions can provide string keys. Those definitions were cached unchanged, so commands such as `setgang` could fail to find an otherwise valid grade.

The lint standard currently declares `qbx.math.round` but not `qbx.math.isFinite`, even though `isFinite` is implemented in `modules/lib.lua`. The narrow inline directives prevent those valid calls from failing every repository lint run without changing runtime behavior.

Closes #543

## Validation

- `git diff --check origin/main...HEAD`
- push and pull-request Lua lint workflows completed successfully
- JUnit lint report: 55 passed, 0 failed
- CodeQL completed successfully
- confirmed the pushed branch SHA matches the reviewed local commit
- reviewed all grade lookup and mutation call sites

The repository has no local test harness and no Lua runtime is installed in this workspace, so no package-based Lua checker was invoked.

### DIFF
--- a/server/groups.lua
+++ b/server/groups.lua
@@ -22,6 +22,45 @@ for name in pairs(gangs) do
     end
 end
 
+---@param grade unknown
+---@return integer? grade
+local function normalizeGrade(grade)
+    if type(grade) == 'string' then
+        if not grade:match('^%d+$') then return end
+        grade = tonumber(grade)
+    end
+
+    if type(grade) ~= 'number' or grade < 0 or grade % 1 ~= 0 then return end
+
+    return grade
+end
+
+---@param group Job | Gang
+---@return boolean success
+---@return string? message
+local function normalizeGroupGrades(group)
+    if type(group.grades) ~= 'table' then
+        return false, 'Invalid parameter: group grades must be a table.'
+    end
+
+    local grades = {}
+    for grade, data in pairs(group.grades) do
+        local normalizedGrade = normalizeGrade(grade)
+        if not normalizedGrade then
+            return false, ("Invalid grade '%s': grade keys must be non-negative integers."):format(tostring(grade))
+        end
+
+        if grades[normalizedGrade] ~= nil then
+            return false, ("Invalid grade '%s': multiple grade keys resolve to the same integer."):format(tostring(grade))
+        end
+
+        grades[normalizedGrade] = data
+    end
+
+    group.grades = grades
+    return true
+end
+
 --- Removes any quotes to ensure functionality
 ---@param str string
 ---@return string
@@ -148,6 +187,9 @@ function CreateJob(jobName, job, commitToFile)
         return false, "Invalid parameter: job must be a table."
     end
 
+    local gradesValid, gradesError = normalizeGroupGrades(job)
+    if not gradesValid then return false, gradesError end
+
     -- Store the job data
     jobs[jobName] = job
 
@@ -220,13 +262,31 @@ exports('RemoveJob', RemoveJob)
 ---Adds or overwrites gangs in shared/gangs.lua
 ---@param newGangs table<string, Gang>
 ---@param commitToFile boolean Whether to commit the gang data to the shared file.
+---@return boolean success
+---@return string? message
 function CreateGangs(newGangs, commitToFile)
+    if type(newGangs) ~= 'table' then
+        return false, 'Invalid parameter: newGangs must be a table.'
+    end
+
+    for gangName, gang in pairs(newGangs) do
+        if type(gang) ~= 'table' then
+            return false, ("Invalid parameter: gang '%s' must be a table."):format(tostring(gangName))
+        end
+
+        local gradesValid, gradesError = normalizeGroupGrades(gang)
+        if not gradesValid then
+            return false, ("Gang '%s': %s"):format(tostring(gangName), gradesError)
+        end
+    end
+
     for gangName, gang in pairs(newGangs) do
         gangs[gangName] = gang
         notifyGroupUpdate('Gang', gangName)
     end
 
     commitGroupToFile('Gang', commitToFile)
+    return true
 end
 
 exports('CreateGangs', CreateGangs)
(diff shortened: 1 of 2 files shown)


server/groups.lua
```diff
@@ -22,6 +22,45 @@ for name in pairs(gangs) do
     end
 end
 
+---@param grade unknown
+---@return integer? grade
+local function normalizeGrade(grade)
```
Kenshiin13, line 27:
```lua
---@param grade unknown
---@return integer?
local function normalizeGrade(grade)
    grade = tonumber(grade)
    if math.type(grade) ~= 'integer' or grade < 0 then return end
    return grade
end
```

Kenshiin13 (changes requested):

server/groups.lua
```diff
@@ -22,6 +22,45 @@ for name in pairs(gangs) do
...
+    if type(grade) == 'string' then
+        if not grade:match('^%d+$') then return end
+        grade = tonumber(grade)
+    end
+
+    if type(grade) ~= 'number' or grade < 0 or grade % 1 ~= 0 then return end
```
pushkart2, line 33: **`normalizeGrade` accepts integral floats with no int64 representation, which can poison the cache.** `grade:match('^%d+$')` allows arbitrarily long digit strings, and in Lua 5.4 `tonumber('99999999999999999999')` returns the float `1e20`, for which `1e20 % 1 == 0` — so this integrality check passes and `grades[1e20]` is stored as a float key, contradicting the `---@return integer?` contract.

The job is cached and broadcast *before* `commitToFile`, and `convertGroupsToPlainText` then hits `string.format("[%d]", 1e20)` which raises `number has no integer representation`. Because the poisoned entry stays in the cache and the file writer iterates all jobs, every later `commitToFile` re-raises until restart.

This is exactly the invalid-key class the PR set out to reject. Letting `math.tointeger` do both checks fixes it cleanly:
```lua
local function normalizeGrade(grade)
    if type(grade) == 'string' then
        if not grade:match('^%d+$') then return end
        grade = tonumber(grade)
    end
    if type(grade) ~= 'number' then return end
    grade = math.tointeger(grade)
    if not grade or grade < 0 then return end
    return grade
end
```

server/groups.lua
```diff
@@ -335,7 +395,12 @@ local function upsertJobGrade(name, grade, data, commitToFile)
         lib.print.error('Job must exist to edit grades. Not found:', name)
         return
     end
-    jobs[name].grades[grade] = data
+    local normalizedGrade = normalizeGrade(grade)
```
pushkart2, line 398: **Normalization asymmetry: `removeJobGrade`/`removeGangGrade` still use the raw key** (`jobs[name].grades[grade] = nil` at lines 439/454). With this PR, `UpsertJobGrade('police', '5', data)` stores integer key `5`, but `RemoveJobGrade('police', '5')` deletes string key `'5'` — a silent no-op: `notifyGroupUpdate` still fires and `commitToFile` rewrites the shared file **still containing the grade**. Before the PR both paths used the raw key, so add/remove was at least symmetric for QB-style string callers.

Concrete failure: an admin/job-manager resource deletes a boss grade it created with a string key — the deletion looks successful but the grade (including `isboss`) persists and stays assignable. Suggest running the same `normalizeGrade` in both remove functions.

server/groups.lua
```diff
@@ -220,13 +262,31 @@ exports('RemoveJob', RemoveJob)
...
+    if type(newGangs) ~= 'table' then
+        return false, 'Invalid parameter: newGangs must be a table.'
+    end
+
+    for gangName, gang in pairs(newGangs) do
+        if type(gang) ~= 'table' then
+            return false, ("Invalid parameter: gang '%s' must be a table."):format(tostring(gangName))
+        end
+
+        local gradesValid, gradesError = normalizeGroupGrades(gang)
```
pushkart2, line 277: Two consistency points on the new gang validation:

1. **`CreateGangs` is all-or-nothing while `CreateJobs` is per-item.** One invalid gang silently drops *every* gang in the batch, whereas `CreateJobs` creates the valid ones and reports the failures. No existing caller reads the new return values, so a mixed batch just vanishes with no log. Consider matching `CreateJobs`' per-item skip+report semantics.
2. **The deprecated QB bridge swallows the new failures.** In `bridge/qb/server/functions.lua`, `addJob`/`addJobs`/`addGang`/`addGangs`/`updateJob`/`updateGang` call `Create*` and unconditionally `return true, 'success'` (e.g. lines 269-270) — the QB-style callers most likely to trip the new grade-key rejection are told success while the group was dropped. Suggest propagating the `Create*` result (or at least `lib.print.error` on failure) through those wrappers.

pushkart2 (commented):
**Automated review of the grade-key normalization.** The core fix is correct: grade `0` survives the truthiness checks (0 is truthy in Lua), integral-float keys normalize via Lua 5.4 table semantics, `CreateJobs` delegates to the now-normalized `CreateJob`, and no sibling resource in the org indexes grades with string keys (verified by grep). Three consistency gaps are called out inline below.

_Minor note (no change strictly required):_ the read side of the QB bridge (`bridge/qb/server/main.lua` exposes `GetJobs()`/`GetGangs()` by reference) still returns integer grade keys, so a third-party QB script that both writes a job via `AddJob('x', {grades={['0']=...}})` and later reads `Shared.Jobs.x.grades['0']` will now get nil. This is the open question from the end of #543 — worth an explicit answer/decision on the PR, though blast radius is limited (no org resource does string-key grade reads).

